PR #5, opened 2021-07-31
Add isoeditor from assisted-service to create minimal iso templates

carbonin:
This PR makes the image service create the minimal iso template image for each configured version on startup.
It also makes these images accessible using a new query parameter, `type`.

The first several commits are the history of the `isoeditor` and `isoutil` packages from openshift/assisted-service. These are included so that we can reference the decisions we made about the implementation details of this code from this repo in the future.

The later commits detail how I transformed the existing packages into something usable by this project.
At a high level, what was done was:

- Removed the code related to creating the cluster image (as this will be done using a "streaming" approach)
- Removed the interface and struct for isoutil and moved the resulting functions into isoeditor
- Used isoeditor from the existing `imagestore` package to create the minimal template images in `Populate`
- Added a query parameter to the API to indicate if a `full` or `minimal` image should be downloaded


carbonin:
Tests are failing because we don't have `genisoimage` in the test container.
Going to leave solving that until Monday.

filanov:
suggesting to follow the same template format that we have in assisted-service https://github.com/openshift/assisted-service/blob/master/docs/pull_request_template.md

filanov:
Why are you writing the handlers manually and not using go-swagger?

carbonin:
> Why are you writing the handlers manually and not using go-swagger?

I feel like swagger's main use cases are:
1. Clearly document a complex API
2. Allow for generating clients in multiple languages
3. Create server-side structures for complicated data in requests

We don't have any of these problems. There's only one API endpoint, we're not going to have any programmatic API access and we only have a few query parameters. I don't see what we would gain by adding it.

carbonin:
Also, @djzager if we let prow merge this, is it going to squash all these commits?
Is there some way to prevent that? The whole reason I went through the trouble of extracting the history was so that we could reference it later.

carbonin:
/hold

Until we can ensure the commit history is preserved.

filanov:
> > Why are you writing the handlers manually and not using go-swagger?
> 
> I feel like swagger's main use cases are:
> 
> 1. Clearly document a complex API
> 2. Allow for generating clients in multiple languages
> 3. Create server-side structures for complicated data in requests
> 
> We don't have any of these problems. There's only one API endpoint, we're not going to have any programmatic API access and we only have a few query parameters. I don't see what we would gain by adding it.

For now we have only one API, maybe it will remain the same and maybe not.
If we will add the range support (don't remember the correct name, download from offset) then we will have more parameters etc that can be handled by swagger. 
But if this api is not directly called by the user then maybe it's fine.

carbonin:
> For now we have only one API, maybe it will remain the same and maybe not.

If it becomes necessary, then we can definitely add it, but it doesn't feel worth the complexity at this time.

> If we will add the range support (don't remember the correct name, download from offset) then we will have more parameters etc that can be handled by swagger.

This is not something swagger will solve for us, but it is already being handled correctly in this repo using [http.ServeContent](https://github.com/openshift/assisted-image-service/blob/d30452bcb1a7be440adcc47f0289723220720e65/internal/handlers/images.go#L90)

carbonin:
Yup, [this repo is set to squash when prow merges](https://github.com/openshift/release/blob/9d376b2609fb40ad9a08be8c6c5a949937984f61/core-services/prow/02_config/openshift/assisted-image-service/_prowconfig.yaml#L3) so we'll have to merge this one manually when the tests start passing.

djzager (approved):

PR #8, opened 2021-08-09
MGMT-7256 Add ISO stream customization

carbonin:
Query the ignition content from the configured assisted service and embed it into the appropriate RHCOS base iso.

The embed area location and size is determined by inspecting the base iso directly rather than putting that information in the iso system area just to read it out later.

This is also using the v1 api currently as the ignition download endpoint for infraenv has not been created yet and it's unclear when the infraenv is going to be created.

The iso streaming implementation is taken from the POC work done by myself and @zaneb in https://github.com/zaneb/iso-stream.

This doesn't quite complete https://issues.redhat.com/browse/MGMT-7256 yet as we still need support for a few more things. Specifically:
- Querying the minimal iso initrd from assisted service (https://github.com/openshift/assisted-service/pull/2342)
- Embedding the initrd into the iso stream
- Handling authentication tokens


flaper87:
> and it's unclear when the infraenv is going to be created.

Can you elaborate a bit on this? Do you mean that it's unclear when the endpoint will be added? Or are there some parts of the image workflow creation that are still unclear?

carbonin:
> Can you elaborate a bit on this? Do you mean that it's unclear when the endpoint will be added? Or are there some parts of the image workflow creation that are still unclear?

When we fully cut over to v2 then the infraenv will be created by the user prior to trying to download the image, but we're not there yet. I was hoping that for compatibility the infraenv would have been created at cluster creation time in the v1 API, but that isn't currently the case. Currently the infraenv is only created in the generate iso endpoint which we're obviously avoiding here.

So to get this to work currently we either need to be using a fully v2 flow or generate the iso before attempting to stream the iso using this service.

All that said, there's no v2 API to download the discovery ignition yet, so this one is our only option for now.

> Is iso-stream not ready to be depended on? I'm a bit unclear on what bits you are taking from Zane's fork, vs what it's possible to use as a go dependency.

It was never really intended as a standalone module. I only created it as a proof of concept.
I rearranged it a bit, but most of `stream.go` and all of the `overlay` package is from Zane's fork.
I spoke with him about how to organize this so we could both benefit from the work and the resolution was to organize the packages in this project so they could be depended on without also utilizing the http server aspects.

flaper87:
Thank you for the details, @carbonin 

> It was never really intended as a standalone module

I was not aware of this, it makes much more sense now. Thanks

internal/handlers/images.go
```diff
@@ -1,34 +1,45 @@
...
+	ImageStore            imagestore.ImageStore
+	AssistedServiceScheme string
+	AssistedServiceHost   string
+	GenerateImageStream   isoeditor.StreamGeneratorFunc
 }
 
 var _ http.Handler = &ImageHandler{}
 
-var clusterRegexp = regexp.MustCompile(`/images/.+`)
+var clusterRegexp = regexp.MustCompile(`^/images/(.+)`)
```
djzager, line 32: Nice change.

internal/handlers/images.go
```diff
@@ -94,15 +112,73 @@ func (h *ImageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+func (h *ImageHandler) imageStreamForID(imageID string, baseISOPath string) (io.ReadSeeker, error) {
+	ignition, err := h.ignitionContent(imageID)
+	if err != nil {
+		return nil, err
+	}
+	return h.GenerateImageStream(baseISOPath, ignition)
+}
+
+func (h *ImageHandler) ignitionContent(imageID string) ([]byte, error) {
```
djzager, line 137: Is this something that should be done outside the handler? I had expected `imageStreamForID` to call something internal to grab it from the assisted-service.

main.go
```diff
@@ -14,11 +14,12 @@ import (
 )
 
 var Options struct {
-	AssistedServiceURL string `envconfig:"ASSISTED_SERVICE_URL" default:"http://assisted-service:8080"`
-	Port               string `envconfig:"PORT" default:"8080"`
-	HTTPSKeyFile       string `envconfig:"HTTPS_KEY_FILE"`
-	HTTPSCertFile      string `envconfig:"HTTPS_CERT_FILE"`
-	DataDir            string `envconfig:"DATA_DIR"`
+	AssistedServiceScheme string `envconfig:"ASSISTED_SERVICE_SCHEME"`
```
djzager, line 17: I guess it's easier to split it here then via regex later?

djzager: or via the urllib tools.

pkg/isoeditor/stream.go
```diff
@@ -0,0 +1,44 @@
+package isoeditor
```
djzager, line 1: Starting from here in the morning. Need to mark my place :sunglasses:

djzager (commented):
A few comments while I'm only halfway through.

main.go
carbonin: Yeah, I'm actually split on this. Theoretically I could make this entire thing some kind of url template.

Then it would go something like:
1. Replace `{image_id}` in the template with our request it
2. Parse the url out
3. Add other info to url (auth, etc.)
4. Use that for request

The more I think about it the more I want to go this direction. Do you think it's worth doing that in this patch now?

internal/handlers/images.go
carbonin: I'm not sure, I was thinking about creating some kind of "client" for this, but didn't feel like it was worth it just yet. Maybe I'll refactor when I add the minimal iso and initrd query.

internal/handlers/images.go
carbonin: This was actually breaking things so I created a separate commit and tests for this.

djzager (approved):
/lgtm

main.go
carbonin: That all said I think I misunderstood your initial question.
I split out the host and scheme so that I didn't have to parse them out and then recombine them later.

I could do it that way though if you would rather prioritize reducing the number of env vars. I don't feel super strongly about it either way.

PR #9, opened 2021-08-11
MGMT-7499 Take an auth token as a query parameter and pass it to assisted

carbonin:
We will either pass this token in another query parameter or as a header
depending on the value of the `REQUEST_AUTH_TYPE` env var.

https://issues.redhat.com/browse/MGMT-7499

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ skipper make test
 - `PORT` - Service listen port
 - `HTTPS_KEY_FILE` - tls key file path
 - `HTTPS_CERT_FILE` - tls cert file path
-- `ASSISTED_SERVICE_URL` - URL to use to query assisted service for image information
+- `ASSISTED_SERVICE_SCHEME` - protocol to use to query assisted service for image information
+- `ASSISTED_SERVICE_HOST` - host to use to query assisted service for image information
+- `REQUEST_AUTH_TYPE` - determines how the auth token should be passed to the assisted service - either `header` or `param`
+  - For `header` a header of the form `Authorization: Bearer <token>` is used
+  - For `param` an `api_key=<token>` query parameter is added to the URL
 
 Example `RHCOS_VERSIONS`:
 ```json
@@ -57,6 +61,7 @@ Downloads the RHCOS image for the specified image ID.
 
 `version`: indicates the version of the RHCOS base image to use (must match a key in `RHCOS_VERSIONS`)
 `type`: `full` to download the ISO including the rootfs, `minimal` to download the iso without the rootfs
+`api_key`: the api token to pass through to the assisted service calls if authentication is required
 
 ### `GET /health`
 
--- a/internal/handlers/images.go
+++ b/internal/handlers/images.go
@@ -20,11 +20,17 @@ import (
 	stdmiddleware "github.com/slok/go-http-metrics/middleware/std"
 )
 
+const (
+	RequestAuthTypeHeader = "header"
+	RequestAuthTypeParam  = "param"
+)
+
 type ImageHandler struct {
 	ImageStore            imagestore.ImageStore
 	AssistedServiceScheme string
 	AssistedServiceHost   string
 	GenerateImageStream   isoeditor.StreamGeneratorFunc
+	RequestAuthType       string
 }
 
 var _ http.Handler = &ImageHandler{}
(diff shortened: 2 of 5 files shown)


omertuc:
Hey @carbonin , I wanted to review this PR but I know nothing about the architecture of the new image service. Where does it run? How does it communicate with the Assisted Service (or vice versa?)?

Is there any design doc / overview for this?

carbonin:
The enhancement doc is here https://github.com/openshift/assisted-service/blob/master/docs/enhancements/image-handling-service.md

Other than that we just have the epic, which probably isn't as helpful as the doc: https://issues.redhat.com/browse/MGMT-3934

carbonin:
@omertuc to answer your other questions directly though:

> Where does it run?

Theoretically anywhere. With this patch I ran it locally with podman, but pointed it at the assisted service integration environment. I was able to create a cluster in the cloud env and use this service to create the image. In practice it will eventually run as a deployment in the same namespace as assisted service.

> How does it communicate with the Assisted Service (or vice versa?)?

The assisted service host and scheme are set using env vars. Thanks for reminding me that I need to update the README :sweat_smile:

main.go
```diff
@@ -16,10 +16,11 @@ import (
 var Options struct {
 	AssistedServiceScheme string `envconfig:"ASSISTED_SERVICE_SCHEME"`
 	AssistedServiceHost   string `envconfig:"ASSISTED_SERVICE_HOST"`
-	Port                  string `envconfig:"PORT" default:"8080"`
+	DataDir               string `envconfig:"DATA_DIR"`
```
omertuc, line 19: What is that?

main.go
```diff
@@ -16,10 +16,11 @@ import (
...
 	HTTPSKeyFile          string `envconfig:"HTTPS_KEY_FILE"`
 	HTTPSCertFile         string `envconfig:"HTTPS_CERT_FILE"`
-	DataDir               string `envconfig:"DATA_DIR"`
+	Port                  string `envconfig:"PORT" default:"8080"`
```
omertuc, line 22: Is that the image service port or service port? (I could figure that out from the rest of the code obviously, just saying it might be worth to give this a better name)

omertuc: Can the service port even be specified?

README.md
```diff
@@ -27,7 +27,9 @@ skipper make test
 - `PORT` - Service listen port
 - `HTTPS_KEY_FILE` - tls key file path
 - `HTTPS_CERT_FILE` - tls cert file path
-- `ASSISTED_SERVICE_URL` - URL to use to query assisted service for image information
+- `ASSISTED_SERVICE_SCHEME` - protocol to use to query assisted service for image information
+- `ASSISTED_SERVICE_HOST` - host to use to query assisted service for image information
+- `REQUEST_AUTH_TYPE` - determines how the auth token should be passed to the assisted service - either `header` or `param`
```
omertuc, line 32: Maybe add a word or two about the distinction between them? "header" being HTTP Authorization header bearer token, "param" being a URL query parameter

main.go
carbonin: This was just a reordering.

The data directory is where we'll store the template images locally and the directory we use for scratch space while we're creating the minimal iso template.

main.go
carbonin: This is the port the image service binds to.
The port for assisted service can be specified in the `ASSISTED_SERVICE_HOST` var.

main.go
omertuc: Yeah wanted to delete the comment after realizing it was juts reorder. Thanks for the explanation

internal/handlers/images.go
```diff
@@ -67,39 +74,10 @@ func (h *ImageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	imageType := params.Get("type")
-	if imageType == "" {
-		w.WriteHeader(http.StatusBadRequest)
-		_, err = w.Write([]byte("'type' parameter required"))
-		if err != nil {
-			log.Errorf("Failed to write response: %v\n", err)
-		}
-		return
-	} else if imageType != imagestore.ImageTypeFull && imageType != imagestore.ImageTypeMinimal {
+	version, imageType, apiKey, err := h.getQueryParams(r.URL.Query())
```
omertuc, line 77: `getQueryParams` name mislead me a bit, maybe `parseQueryParams`? `getQueryParams` sounds a bit like you're constructing query params for a request you want to make to a different service

main.go
omertuc: Alright. Maybe rename this field to `LISTEN_PORT` and add documentation for `ASSISTED_SERVICE_HOST` that it expects a `HOST:PORT` format

carbonin: I'd prefer not to do that in this PR.

I can do it in a follow up, or you can open that PR if you have the time.

omertuc: Sure will do

PR #10, opened 2021-08-12
Add pull request template

carbonin:
This is based roughly on the template from assisted-service.

/cc @filanov what do you think?
I didn't think the entire template from assisted service was really necessary since we don't have a ton of complicated CI here and everything should be the same regardless of environment.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,36 @@
+## Description
+<!--
+Include a summary of the change as well as the reasoning behind it including any additional context.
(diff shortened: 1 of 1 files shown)


filanov:
Maybe adding a checklist can help as well so people that will open a PR should at least see it 
```
 Title and description added to both, commit and PR.
 Relevant issues have been associated (see CONTRIBUTING guide)
 Reviewers have been listed
 This change does not require a documentation update (docstring, docs, README, etc)
 Does this change include unit-tests (note that code changes require unit-tests)
 ```

filanov (approved):

PR #17, opened 2021-08-30
Use array format for versions and add image arch support

carbonin:
## Description

This commit moves us from the map versions format inspired by the old
assisted-service `OPENSHIFT_VERSIONS` map to the new format for
`RHCOS_VERSIONS`. The main difference is using an array instead of a
map.

This also adds a cpu_architecture field to each default version entry.

A follow-up PR will add a parameter to the API to specify architecture.

## How was this code tested?

Ran the service locally and tested generating both full and minimal isos for an infra-env in integration.
Both images registered successfully.

## Assignees

/cc @danielerez 
/cc @djzager 

## Links

Related assisted-service PRs:
https://github.com/openshift/assisted-service/pull/2369
https://github.com/openshift/assisted-service/pull/2454

JIRA:
https://issues.redhat.com/browse/MGMT-7453

## Checklist

- [x] ~~This change does not require a documentation update (docstring, `docs`, README, etc)~~ Updated README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,20 +35,26 @@ skipper make test
 
 Example `RHCOS_VERSIONS`:
 ```json
-{
-	"4.6": {
-		"iso_url": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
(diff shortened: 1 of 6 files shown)


djzager (approved):

README.md
```diff
@@ -35,20 +35,26 @@ skipper make test
 
 Example `RHCOS_VERSIONS`:
```
filanov, line 36: Isn't it OS_IMAGES?

PR #18, opened 2021-08-31
Add support for multi-arch base images

carbonin:
## Description

This PR is a follow-up to https://github.com/openshift/assisted-image-service/pull/17 to add an architecture parameter to the API which will help determine which base image to use for a download.

This makes the "key" for the base image the combination of version and architecture. If the version/arch combo in a request is not present in the `RHCOS_VERSIONS` configuration we will return a 400 error.

## How was this code tested?

Ran the service locally with an arm entry in the versions list and successfully downloaded an edited arm iso with an ignition embedded.

## Assignees

/cc @djzager 
/cc @danielerez 

## Links

Resolves https://issues.redhat.com/browse/MGMT-7453

## Checklist

- [x] ~~This change does not require a documentation update (docstring, `docs`, README, etc)~~ README updated


pkg/imagestore/imagestore.go
```diff
@@ -19,28 +19,28 @@ var DefaultVersions = []map[string]string{
...
+		"url":               "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.7/4.7.13/rhcos-4.7.13-x86_64-live.x86_64.iso",
+		"rootfs_url":        "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.7/4.7.13/rhcos-live-rootfs.x86_64.img",
 	},
 	{
 		"openshift_version": "4.8",
 		"cpu_architecture":  "x86_64",
-		"url":               "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-4.8.2-x86_64-live.x86_64.iso",
-		"rootfs_url":        "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img",
+		"url":               "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.8/4.8.2/rhcos-4.8.2-x86_64-live.x86_64.iso",
+		"rootfs_url":        "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img",
```
djzager, line 35: I wonder if we can, or should, fail on startup if we are not given any images to use as a base.

djzager (commented):
Just one comment.

pkg/imagestore/imagestore.go
```diff
@@ -19,28 +19,28 @@ var DefaultVersions = []map[string]string{
 	{
 		"openshift_version": "4.6",
 		"cpu_architecture":  "x86_64",
-		"url":               "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
-		"rootfs_url":        "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-live-rootfs.x86_64.img",
+		"url":               "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
```
danielerez, line 22: nice:) would be clean when using '/arm64/dependencies/' path.

pkg/imagestore/imagestore.go
carbonin: Actually have a ticket open for this already https://issues.redhat.com/browse/MGMT-7735

carbonin: Someone would have to explicitly provide an empty json array as we treat empty string as the default values, so it seems unlikely, but it should be a simple thing to guard against.

Makefile
```diff
@@ -29,6 +29,7 @@ run: certs
 		-e HTTPS_KEY_FILE=/certs/tls.key -e HTTPS_CERT_FILE=/certs/tls.crt \
 		-e ASSISTED_SERVICE_SCHEME=${ASSISTED_SERVICE_SCHEME} -e ASSISTED_SERVICE_HOST=${ASSISTED_SERVICE_HOST} \
 		-e REQUEST_AUTH_TYPE=${REQUEST_AUTH_TYPE} \
+		-e RHCOS_VERSIONS='${RHCOS_VERSIONS}' \
```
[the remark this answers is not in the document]
carbonin: :man_shrugging: Not sure it matters, but this variable isn't being added in this PR so I'd rather change it in a follow-up.

pkg/imagestore/imagestore.go
carbonin: :+1: that's the idea.

pkg/imagestore/imagestore.go
djzager: I guess what I was suggesting was the step further. Since all the deployment mechanisms (dev, cloud, operator) have (or will soon have) the images included...that we could consider not having a default and failing if they aren't specified.

Makefile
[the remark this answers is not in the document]
danielerez: Sure, can be renamed later on if we want. The logic behind using a generic OS naming was just to avoid being too specific with RHCOS, as we might support other distributions.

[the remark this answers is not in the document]
carbonin: Yup, I totally understand the reasoning, I just don't think the change belongs in this PR.

PR #19, opened 2021-09-02
feat: support self-signed certificates on assisted-service

djzager:
## Description

The assisted-service may likely be using self-signed certificates. This
change makes it possible to start the imagehandler with a cabundle to
make trust between the image-service and assisted-service possible.

Additionally, we break up the handling of `err != nil` to ensure we
don't panic because response != 2xx but err was nil.

## How was this code tested?

dev-scripts environment

## Assignees

/cc @carbonin 
/cc @

## Links
https://github.com/openshift/assisted-service/pull/2480

## Checklist

### DIFF
--- a/internal/handlers/images.go
+++ b/internal/handlers/images.go
@@ -3,8 +3,11 @@ package handlers
 import (
 	"bytes"
 	"compress/gzip"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -35,6 +38,7 @@ type ImageHandler struct {
 	AssistedServiceHost   string
 	GenerateImageStream   isoeditor.StreamGeneratorFunc
 	RequestAuthType       string
+	Client                *http.Client
 }
 
 var _ http.Handler = &ImageHandler{}
@@ -49,7 +53,7 @@ func parseImageID(path string) (string, error) {
 	return match[1], nil
 }
 
-func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedServiceHost, requestAuthType string) http.Handler {
+func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedServiceHost, requestAuthType, caCertFile string) http.Handler {
 	metricsConfig := metrics.Config{
 		Prefix:          "assisted_image_service",
 		DurationBuckets: []float64{.1, 1, 10, 50, 100, 300, 600},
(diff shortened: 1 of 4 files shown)


internal/handlers/images.go
```diff
@@ -59,12 +63,30 @@ func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedSe
...
 	})
 
+	client := http.DefaultClient
+	if caCertFile != "" {
+		caCert, err := ioutil.ReadFile(caCertFile)
+		if err != nil {
+			log.Fatalf("Error opening cert file %s, %s", caCertFile, err)
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
```
flaper87, line 73: It looks like [AppendCertsFromPEM](https://pkg.go.dev/crypto/x509#CertPool.AppendCertsFromPEM) returns a boolean to indicate whether it succeeded or not.

What do you think about consuming the return value and print a warning/error in case of failure?

flaper87 (changes requested):
suggestion in-line. Otherwise, lgtm

internal/handlers/images.go
djzager: My thinking is that if we can't add the cert, we should fail since there should only ever be one that we are adding **and** we obviously need it.

djzager:
I wanted to rerun all of the unit tests against a TLS server that is provided by `gttp` via `NewTLSServer()`. It isn't immediately obvious how to structure our tests such that any new tests could run against both an http and https server.
Think we should create a follow-up task so we can prioritize this effort.

internal/handlers/images.go
flaper87: > My thinking is that if we can't add the cert, we should fail since there should only ever be one that we are adding and we obviously need it.

Failing in this case makes sense. Are you going to do the change in this PR?

djzager: I'm glad you asked that because I thought I had done it already.

internal/handlers/images.go
```diff
@@ -59,12 +63,31 @@ func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedSe
 		Recorder: metrics.NewRecorder(metricsConfig),
 	})
 
+	client := http.DefaultClient
```
carbonin, line 66: `DefaultClient` is actually an instantiated client. I think it would be better to create a new client rather than editing what is effectively global state.

carbonin (commented):
Is this required for any current flow or is this mostly just for testing?

Specifically I don't think this needs to go in for https://github.com/openshift/assisted-service/pull/2480 to work, right?

carbonin:
Answered my own question https://github.com/openshift/assisted-service/pull/2480#discussion_r703491567

Service serving certs are not really self-signed, they're just not signed by a public root CA. OpenShift has an internal CA that it uses to sign these certs which is generally accepted on the platform (as opposed to self-signed, which is a bad idea in prod).

Mentioning this because "self-signed" in the title made me think this was a convenience thing for testing when really this is required.

internal/handlers/images.go
```diff
@@ -59,12 +63,31 @@ func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedSe
...
+		if err != nil {
+			log.Fatalf("Error opening cert file %s, %s", caCertFile, err)
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		t := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs:    caCertPool,
+				MinVersion: tls.VersionTLS12,
```
carbonin, line 78: What kind of errors were you seeing that prompted specifying a minimum version?

djzager: The linter was not happy with me. This was the minimum version that made it stop complaining.

carbonin: Fair enough

flaper87 (approved):
:pizza:

PR #21, opened 2021-09-09
MGMT-7754: Support iso creation with EFI boot only

danielerez:
## Description
As the aarch64 rhcos image[*] contains a single boot image (efiboot.img),
added proper support to isoutil for creating an ISO with only that boot platform.
This should facilitate the creation of a minimal ISO template for arm64 CPU architecture.

Note: same as done in assisted-service - https://github.com/openshift/assisted-service/pull/2563

[*] https://mirror.openshift.com/pub/openshift-v4/aarch64/dependencies/rhcos/pre-release/4.9.0-fc.1/rhcos-4.9.0-fc.1-aarch64-live.aarch64.iso

## How was this code tested?
Tested same change in assisted-service.

## Assignees
/cc @carbonin 

## Checklist

### DIFF
--- a/pkg/isoeditor/isoutil.go
+++ b/pkg/isoeditor/isoutil.go
@@ -146,6 +146,27 @@ func Create(outPath string, workDir string, volumeLabel string) error {
 				},
 			},
 		}
+	} else if exists, _ := fileExists(filepath.Join(workDir, "images/efiboot.img")); exists {
+		// Creating an ISO with EFI boot only
+		efiSectors, err := efiLoadSectors(workDir)
+		if err != nil {
+			return err
+		}
+		if exists, _ := fileExists(filepath.Join(workDir, "boot.catalog")); !exists {
+			return fmt.Errorf("missing boot.catalog file")
+		}
+		options.ElTorito = &iso9660.ElTorito{
+			BootCatalog:     "boot.catalog",
+			HideBootCatalog: true,
+			Entries: []*iso9660.ElToritoEntry{
+				{
+					Platform:  iso9660.EFI,
+					Emulation: iso9660.NoEmulation,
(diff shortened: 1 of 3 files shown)


pkg/isoeditor/isoutil.go
```diff
@@ -146,6 +146,24 @@ func Create(outPath string, workDir string, volumeLabel string) error {
...
 			},
 		}
+	} else if exists, _ := fileExists("images/efiboot.img"); exists {
+		// Creating an ISO with EFI boot only
+		efiSectors, err := efiLoadSectors(workDir)
+		if err != nil {
+			return err
+		}
+		options.ElTorito = &iso9660.ElTorito{
+			BootCatalog:     "boot.catalog",
```
carbonin, line 156: Should we check that this file exists? It doesn't in the x86 isos.

carbonin (commented):
I don't think there's a way to validate that the boot entries were created correctly, but can you add a test that at least runs through the new branch of code?

pkg/isoeditor/isoutil.go
carbonin: I know it won't affect them. My point is just that it doesn't exist in some cases so we should check for it.

danielerez: Yeah, so I've meant that it should always exist if the iso has only one boot file afaik. But can add a check just in case.

danielerez: Added.

danielerez:
> I don't think there's a way to validate that the boot entries were created correctly, but can you add a test that at least runs through the new branch of code?

Sure, added some tests.

carbonin (approved):

PR #23, opened 2021-09-09
Align type values with those from assisted-service

carbonin:
## Description

When constructing the download URL from assisted-service it will be
better to use the image type directly from that API rather than
translating it.

Specifically changes "full" to "full-iso" and "minimal" to "minimal-iso"

## How was this code tested?

Unit tests cover this rather well.

## Assignees

/cc @djzager 
/cc @flaper87 

## Links

Related to https://issues.redhat.com/browse/MGMT-7258

## Checklist

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -55,8 +55,8 @@ type rhcosStore struct {
 }
 
 const (
-	ImageTypeFull    = "full"
-	ImageTypeMinimal = "minimal"
+	ImageTypeFull    = "full-iso"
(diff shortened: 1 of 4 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -55,8 +55,8 @@ type rhcosStore struct {
 }
 
 const (
-	ImageTypeFull    = "full"
-	ImageTypeMinimal = "minimal"
+	ImageTypeFull    = "full-iso"
```
djzager, line 58: I agree. It got me a few times.

djzager (approved):
:cookie:

PR #25, opened 2021-09-22
MGMT-7762: Switch to stream8

flaper87:
## Description
The original plan was to move all images to ubi8. This is not possible due to the lack
of some packages that are needed for other projects. We are now going to switch all images
to stream8 with the hope that the consistency accross repos will prevent (or help) with
debugging current/future issues in CI.

The goal is to keep component's builds as consistent as possible in the channels we are
releasing them on

Signed-off-by: Flavio Percoco <flavio@redhat.com>

## How was this code tested?
It wasn't! Relying on CI for this 

## Assignees

/cc @carbonin 
/cc @djzager 

## Checklist

### DIFF
--- a/Dockerfile.image-service
+++ b/Dockerfile.image-service
@@ -6,7 +6,7 @@ RUN go mod download
 COPY . .
 RUN CGO_ENABLED=0 GOFLAGS="" GO111MODULE=on go build -o /assisted-image-service main.go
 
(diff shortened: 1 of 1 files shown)


carbonin:
Looks like it only increases the image size by 30M, I guess minimal isn't so minimal :laughing: 

/lgtm

PR #27, opened 2021-09-27
Limit requests in flight for image downloads

carbonin:
## Description

When testing large numbers of simultaneous downloads (~1000)
some of the requests failed due to hitting an open file limit.

This commit adds an environment variable that allows us to cap the
number of downloads in progress.

## How was this code tested?

Ran the image service locally against a "no-op" assisted service that just returned some content immediately to any request.
I issued 1000 simultaneous requests to download images from the assisted service and ran into issues with local file limits like this:

`{"file":"/home/ncarboni/Source/assisted-image-service/internal/handlers/images.go:138","func":"github.com/openshift/assisted-image-service/internal/handlers.(*ImageHandler).ServeHTTP","level":"error","msg":"Error creating image stream: failed to create overwrite reader for ramdisk: Could not open device data/rhcos-minimal-iso-4.9-x86_64.iso with mode read-only: open data/rhcos-minimal-iso-4.9-x86_64.iso: too many open files\n","time":"2021-09-27T10:23:05-04:00"}`

## Assignees

/cc @djzager 

## Links

https://issues.redhat.com/browse/MGMT-7850

## Checklist

- [ ] ~~This change does not require a documentation update (docstring, `docs`, README, etc)~~ README updated

### DIFF
--- a/internal/handlers/images.go
+++ b/internal/handlers/images.go
@@ -40,6 +40,7 @@ type ImageHandler struct {
 	GenerateImageStream   isoeditor.StreamGeneratorFunc
 	RequestAuthType       string
 	Client                *http.Client
+	sem                   chan struct{}
 }
 
 type imageDownloadParams struct {
@@ -61,7 +62,7 @@ func parseImageID(path string) (string, error) {
 	return match[1], nil
 }
 
-func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedServiceHost, requestAuthType, caCertFile string) http.Handler {
+func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedServiceHost, requestAuthType, caCertFile string, maxRequests int) http.Handler {
 	metricsConfig := metrics.Config{
 		Prefix:          "assisted_image_service",
 		DurationBuckets: []float64{.1, 1, 10, 50, 100, 300, 600},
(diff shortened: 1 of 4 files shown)


internal/handlers/images.go
```diff
@@ -98,12 +99,23 @@ func NewImageHandler(is imagestore.ImageStore, assistedServiceScheme, assistedSe
...
 		RequestAuthType:       requestAuthType,
 		Client:                client,
+		sem:                   make(chan struct{}, maxRequests),
 	}
 
 	return stdmiddleware.Handler("/images/:imageID", mdw, h)
 }
 
 func (h *ImageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	select {
```
carbonin, line 109: Thoughts on extracting this to a function or middleware of some sort?

I didn't on this first pass because I don't think it's going to realistically be reused anywhere.

djzager: I think it's fine where it is.

main.go
```diff
@@ -22,6 +22,7 @@ var Options struct {
 	HTTPSCAFile           string `envconfig:"HTTPS_CA_FILE"`
 	ListenPort            string `envconfig:"LISTEN_PORT" default:"8080"`
 	RequestAuthType       string `envconfig:"REQUEST_AUTH_TYPE"`
+	MaxConcurrentRequests int    `envconfig:"MAX_CONCURRENT_REQUESTS" default:"400"`
```
djzager, line 25: Should we hardcode this for the operator deployment or leave the default?

djzager (approved):

main.go
carbonin: I think it's fine for now. Previously it was not limited at all, now we have a knob.

400 was around the most I could get to succeed running the image service locally on my laptop so I doubt we'll be able to do more in a real deployment.

PR #28, opened 2021-09-28
MGMT-7734: Integration tests for the image-service 

djzager:
## Description

Verify that we can grab a few versions (and architectures) of RHCOS from
the openshift mirror and:
1. Create full-iso, minimal-iso template
2. Stream an RHCOS image with ignition + initrd customizations embedded
   properly

Since this test could take a long time (and longer in the future as we
support more versions and architectures), we'll skip this when run using
`go test -short`. So this change also includes a new make target
`test-integration` to run this test.

## How was this code tested?

By running the tests (`make test-integration`)

## Assignees

/cc @carbonin 
/cc @

## Checklist

- [x] Title and description added to both, commit and PR
- [x] Relevant issues have been associated
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit tests (note that code changes require unit tests)

PR #30, opened 2021-10-05
Bug 2011052: Use the request context when making requests to assisted-service

carbonin:
## Description

Backports #27 #28 and #29

The bug is fixed by #29, but not having the other changes was failing the cherry-pick to 2.4

#27 adds the request max which is effectively reimplemented in #29 
#28 adds some integration tests and makes minimal changes to the service itself

Those two seem safe to backport in order to avoid manually resolving the conflict.

## How was this code tested?

An image with this change was deployed to OpenShift alongside assisted service where I confirmed that canceling a download request on the client killed the running nmstatectl processes in the assisted service pod.

## Assignees

/cc @djzager 
/cc @celebdor 

## Links

https://bugzilla.redhat.com/show_bug.cgi?id=2011052

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,12 @@ lint:
 test:
 	go test ./...
 
+test-short:
+	go test -short ./...
+
+test-integration:
+	go test ./integration_test/...
+
 generate:
 	go generate $(shell go list ./...)
 	$(MAKE) format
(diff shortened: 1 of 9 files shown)


carbonin:
oops target to wrong branch

PR #32, opened 2021-10-06
Bug 2006546: Image service should fail if no RHCOS_IMAGES are provided

pawanpinjarkar:
The image service should fail with an error in the logs when no RHCOS_IMAGES are provided i.e. RHCOS_IMAGES
is set as an empty slice []. When RHCOS_IMAGES is set to an empty slice, no request to download the ISO image will ever succeed.

https://bugzilla.redhat.com/show_bug.cgi?id=2006546

## Assignees

/cc @
/cc @

## Checklist


carbonin:
It would be nice to get all that logic about the default value and checking the length into a function so we can test it.
Is it a valid move to just put it in a function in `main.go` and have a `main_test.go`?

pkg/imagestore/imagestore.go
```diff
@@ -63,7 +63,10 @@ func NewImageStore(ed isoeditor.Editor, dataDir string, versions []map[string]st
 	}, nil
 }
 
-func validateVersions(versions []map[string]string) error {
+func ValidateVersions(versions []map[string]string) error {
```
carbonin, line 66: You could leave this unexported and test `NewImageStore`

pawanpinjarkar:
/hold until I get clarity about duplicate bugs 
https://bugzilla.redhat.com/show_bug.cgi?id=2010276
https://bugzilla.redhat.com/show_bug.cgi?id=2006546

pkg/imagestore/imagestore_test.go
```diff
@@ -175,23 +175,100 @@ var _ = Describe("HaveVersion", func() {
...
+			var err error
+			dataDir, err = os.MkdirTemp("", "imageStoreTest")
+			Expect(err).NotTo(HaveOccurred())
+
+			ctrl = gomock.NewController(GinkgoT())
+			mockEditor = isoeditor.NewMockEditor(ctrl)
+		})
+
+		It("should error when RHCOS_IMAGES are not set i.e. versions is an empty slice", func() {
+			versions = []map[string]string{}
```
carbonin, line 218: If you're redefining versions for each test, then don't mess with the `Describe("HaveVersions"` block, just make these tests a top-level `Describe("NewImageStore"`.

Setting the value of versions here breaks the per-test scope and will mess with anything that expects versions to be defined before the test.

pkg/imagestore/imagestore_test.go
```diff
@@ -175,23 +175,100 @@ var _ = Describe("HaveVersion", func() {
...
+			Expect(store.HaveVersion("4.8", "aarch64")).To(BeFalse())
+		})
 	})
 
-	It("is false for versions that are missing", func() {
-		Expect(store.HaveVersion("4.9", "x86_64")).To(BeFalse())
-		Expect(store.HaveVersion("4.8", "arm64")).To(BeFalse())
-		Expect(store.HaveVersion("4.7", "x86_64")).To(BeFalse())
-		Expect(store.HaveVersion("4.8", "aarch64")).To(BeFalse())
+	Describe("ValidateVersions", func() {
```
carbonin, line 201: This should be `Describe("NewImageStore", func() {` as that's the function being tested.

pkg/imagestore/imagestore_test.go
```diff
@@ -175,23 +175,100 @@ var _ = Describe("HaveVersion", func() {
...
+		var (
+			ctrl       *gomock.Controller
+			mockEditor *isoeditor.MockEditor
```
carbonin, line 204: I don't think you need either of these, just pass `nil` for the editor.

pkg/imagestore/imagestore_test.go
```diff
@@ -175,23 +175,100 @@ var _ = Describe("HaveVersion", func() {
...
+			dataDir    string
+		)
+
+		BeforeEach(func() {
+			var err error
+			dataDir, err = os.MkdirTemp("", "imageStoreTest")
```
carbonin, line 210: I don't think you need this directory either as we're not doing anything with it, but if you are going to create it then you need to clean it up in an `AfterEach` block.

pkg/imagestore/imagestore_test.go
```diff
@@ -195,3 +195,63 @@ var _ = Describe("HaveVersion", func() {
 		Expect(store.HaveVersion("4.8", "aarch64")).To(BeFalse())
 	})
 })
+
+var _ = Describe("NewImageStore", func() {
+
+	It("should error when RHCOS_IMAGES are not set i.e. versions is an empty slice", func() {
```
djzager, line 201: I think you should add one happy path scenario to help us prevent regressions on the NewImageStore specifically.

djzager (commented):
One comment then I think this is good to go.

pkg/imagestore/imagestore_test.go
pawanpinjarkar: Done :heavy_check_mark:

PR #36, opened 2021-10-12
Bug 2012796: Add liveness endpoint

rollandf:
-## Description

- Add new endpoint `live`, that always respond 200.
- Change endpoint `health` to respond 503 until the `Populate` is done.

## How was this code tested?

- Run the service
- Call for /live and check return code is 200 from the start of the service, during and after `Populate`.
- Call for /health during 'Populate' and check return code is 503.
- Call for /health after 'Populate' and check return code is 200.
- 
## Assignees

/cc @carbonin 
/cc @

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,12 @@ Downloads the RHCOS image for the specified image ID.
 
 ### `GET /health`
 
+Returns 503 until the images are downloaded
 Returns 200 if the service is ready to respond to requests
+
+### `GET /live`
+
+Returns 200 if the service is running
 
 ### `GET /metrics`
 
--- a/internal/handlers/images_test.go
+++ b/internal/handlers/images_test.go
@@ -302,3 +302,22 @@ var _ = Describe("parseImageID", func() {
 		Expect(err).To(HaveOccurred())
 	})
 })
+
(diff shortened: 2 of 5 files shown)


internal/handlers/health.go
```diff
@@ -2,10 +2,29 @@ package handlers
...
 
-func NewHealthHandler() http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-	})
+type HealthHandler struct {
+	isEnabled bool
+}
+
+func NewHealthHandler() *HealthHandler {
```
carbonin, line 13: Maybe name this after the probe it's implementing like you did with the other one?
So this would be a `ReadinessHandler` in that case.

internal/handlers/health.go
```diff
@@ -2,10 +2,29 @@ package handlers
...
+	return &HealthHandler{
+		isEnabled: false,
+	}
+}
+
+func (a *HealthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
carbonin, line 19: I feel like we could write a test for this.
What do you think?

PR #38, opened 2021-11-04
MGMT-7897: Rework authentication handling to support image_token

carbonin:
## Description

This commit changes how the image service handles authentication
parameters. Previously an environment variable (`REQUEST_AUTH_TYPE`) was
used to change how the image service would treat the `api_key` query
parameter.

This didn't end up working out as we want to use a different query
parameter name for different auth types and the effect of changing
`REQUEST_AUTH_TYPE` was unclear.

This commit removes `REQUEST_AUTH_TYPE` and implements separate handling
for each authentication strategy. This should clear up how what is passed
to the image service gets used in calls to assisted service.

Specifically the mappings are now:
| image service auth            | assisted service auth     |
|-------------------------------|---------------------------|
| `api_key` query parameter     | `api_key` query parameter |
| `image_token` query parameter | `Image-Token` header      |
| `Authorization` header        | `Authorization` header    |

## How was this code tested?

I ran an image built with this code configured with an assisted-service installation deployed with the changes in https://github.com/openshift/assisted-service/pull/2824 and was able to successfully download an image using a url the new token auth.

## Assignees

/cc @rollandf 
/cc @danielerez 

## Links

https://github.com/openshift/assisted-service/pull/2824
https://issues.redhat.com/browse/MGMT-7897

## Checklist

### DIFF
--- a/internal/handlers/images.go
+++ b/internal/handlers/images.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
@@ -23,11 +22,6 @@ import (
 )
 
 const (
-	RequestAuthTypeHeader = "header"
-	RequestAuthTypeParam  = "param"
-)
-
-const (
 	fileRouteFormat = "/api/assisted-install/v2/infra-envs/%s/downloads/files"
 	defaultArch     = "x86_64"
 )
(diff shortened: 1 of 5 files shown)


internal/handlers/images.go
```diff
@@ -130,7 +121,24 @@ func (h *ImageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	isoReader, err := h.imageStreamForID(r.Context(), clusterID, params)
```
danielerez, line 133: Just wonder if we should keep the imageStreamForID func, as ServeHTTP gets a little big. Or you prefer to avoid logging from another func, etc.

danielerez (commented):
Minor comment.
/lgtm

internal/handlers/images.go
carbonin: :man_shrugging: it didn't seem like it was adding much.

PR #39, opened 2021-11-15
MGMT-8216: Handle persistent storage for the data directory

carbonin:
## Description

This PR makes a few changes to better handle the situation where the data directory is a persistent volume.
Specifically it:
1. Ensures only the files we need are present in the data directory
    - This ensures that we are not leaving files around indefinitely when we upgrade the base image version
2. Regenerates the minimal ISO template file on every run
    - This ensures that we can recover from any possible bugs we might have in the minimal iso creation code without also forcing a version change or any manual intervention.
3. Uses the rhcos release version to name the files
    - This allows us to distinguish between z-stream releases on redeploy. Previously the file name was based only on the X.Y release version.

## How was this code tested?

Ran the code locally using a host-mounted data-directory and observed that existing files were removed and new ones were added on each run.

## Assignees

/cc @filanov 
/cc @danielerez 

## Links

https://issues.redhat.com/browse/MGMT-8216

## Checklist

- [x] ~~This change does not require a documentation update (docstring, `docs`, README, etc)~~ Updated README


pkg/imagestore/imagestore.go
```diff
@@ -154,8 +178,50 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 	return errs.Wait()
 }
 
-func (s *rhcosStore) PathForParams(imageType, version, arch string) string {
-	return filepath.Join(s.dataDir, fmt.Sprintf("rhcos-%s-%s-%s.iso", imageType, version, arch))
+func (s *rhcosStore) PathForParams(imageType, openshiftVersion, arch string) string {
+	var version string
+	for _, entry := range s.versions {
+		if entry["openshift_version"] == openshiftVersion && entry["cpu_architecture"] == arch {
```
danielerez, line 184: do we need fallback to default arch here in case it's missing? i.e. empty means x86. or is it already validated/handled?

pkg/imagestore/imagestore.go
```diff
@@ -154,8 +178,50 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
+	}
+
+	for _, dataDirFile := range dataDirFiles {
+		if !fileExpected(dataDirFile.Name()) {
+			if err := os.RemoveAll(filepath.Join(s.dataDir, dataDirFile.Name())); err != nil {
+				return err
+			}
+		}
+	}
+
```
danielerez, line 223: maybe worth adding a log that folder was successfully cleaned, or/and a warning in case of partially cleaned.

pkg/imagestore/imagestore.go
```diff
@@ -114,15 +134,19 @@ func downloadURLToFile(url string, path string) error {
 }
 
 func (s *rhcosStore) Populate(ctx context.Context) error {
+	if err := s.cleanDataDir(); err != nil {
+		return err
```
flaper87, line 138: Will this make the service fail? Or is it just a runtime error? (not super familiar with the code base, trying to learn)

A more philosophical question:

Should the service be considered healthy if the `Populate` function fails (whether it's removing or writing files)? (maybe this is already handled)

flaper87 (commented):
Thinking outloud: Wonder how this will work once the images are part of the OCP release payload. I guess lots of this logic will stay somewhat the same.

lg, I will let final approval to others.

pkg/imagestore/imagestore.go
```diff
@@ -132,19 +156,19 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 					return fmt.Errorf("failed to download %s: %v", url, err)
 				}
 
-				log.Infof("Finished downloading for %s-%s", openshiftVersion, arch)
+				log.Infof("Finished downloading for %s-%s", imageVersion, arch)
```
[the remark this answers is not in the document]
carbonin: I don't know, each of the "actual" version numbers starts with the ocp major/minor so I don't think it will be too confusing.

That said, it's just a log. I don't have strong opinions on it.

pkg/imagestore/imagestore.go
carbonin: > Will this make the service fail?

Right now, [yes](https://github.com/openshift/assisted-image-service/blob/24cda2dcc77cbf2dc0bcb601b41b8391d5cceccc/main.go#L56-L59). I struggled with this one for a while.

> Should the service be considered healthy if the Populate function fails

If the deletes fail ... maybe, but we cannot consider it healthy if the writes don't succeed.

pkg/imagestore/imagestore.go
```diff
@@ -114,15 +134,19 @@ func downloadURLToFile(url string, path string) error {
...
+		return err
+	}
+
 	errs, _ := errgroup.WithContext(ctx)
 
 	for i := range s.versions {
 		version := s.versions[i]
 		errs.Go(func() error {
-			openshiftVersion := version["openshift_version"]
+			imageVersion := version["version"]
```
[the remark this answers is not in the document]
carbonin: Yeah, I'll update this.

pkg/imagestore/imagestore.go
filanov: we plan to use fallback for day2, not sure if this api knows if it's for day1 or day2

pkg/imagestore/imagestore.go
```diff
@@ -154,8 +178,50 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
+}
+
+func (s *rhcosStore) cleanDataDir() error {
+	var expectedFiles []string
+	for _, version := range s.versions {
+		// Only add full isos here as we want to regenerate the minimal image on each deploy
+		expectedFiles = append(expectedFiles, isoFileName(ImageTypeFull, version["version"], version["cpu_architecture"]))
+	}
+
+	fileExpected := func(name string) bool {
```
filanov, line 202: not a must but you can use https://github.com/thoas/go-funk

pkg/imagestore/imagestore.go
filanov: I think that it should fail if we have issues with the storage we should not hide this issue

pkg/imagestore/imagestore.go
carbonin: It does not, but we create the url in assisted-service so we should be able to do whatever defaulting we need on that side.

pkg/imagestore/imagestore.go
carbonin: I think we'll keep this as is then. I don't expect to have too many problems with this.

pkg/imagestore/imagestore.go
carbonin: Added

filanov (approved):

PR #42, opened 2021-11-17
Add the ability to customise PXE files

zaneb:
In addition to the existing `NewRHCOSStreamReader()` function to generate a streamable customised ISO file, add a new `NewInitRamFSStreamReader()` function to do equivalent customisation of an initramfs file for the purposes of PXE booting.

The [initramfs format](https://www.kernel.org/doc/html/latest/driver-api/early-userspace/buffer-format.html) consists of a sequence of gzipped cpio archives (each archive is gzipped separately), so the customisation can be performed by simply appending a new archive at the end of the stream. The archive itself is identical to the one used to customise the ISO.

### DIFF
--- a/pkg/isoeditor/initramfs.go
+++ b/pkg/isoeditor/initramfs.go
@@ -1,0 +1,27 @@
+package isoeditor
+
+import (
+	"io"
+	"os"
+
+	"github.com/openshift/assisted-image-service/pkg/overlay"
+	"github.com/pkg/errors"
+)
+
+func NewInitRamFSStreamReader(irfsPath string, ignitionContent *IgnitionContent) (io.ReadSeeker, error) {
+	irfsReader, err := os.Open(irfsPath)
+	if err != nil {
+		return nil, err
+	}
+
+	ignitionReader, err := ignitionContent.Archive()
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := overlay.NewAppendReader(irfsReader, ignitionReader)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create append reader for ignition")
+	}
+	return r, nil
+}
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -59,6 +59,25 @@ func NewOverlayReader(base io.ReadSeeker, overlay Overlay) (io.ReadSeeker, error
 	return newReader(base, overlay, length)
 }
 
+func NewAppendReader(base io.ReadSeeker, reader io.ReadSeeker) (io.ReadSeeker, error) {
+	length, err := base.Seek(0, io.SeekEnd)
+	if err != nil {
+		return nil, err
+	}
+
+	appendLength, err := reader.Seek(0, io.SeekEnd)
+	if err != nil {
+		return nil, err
+	}
+
+	overlay := Overlay{
+		Reader: reader,
+		Offset: length,
+		Length: appendLength,
+	}
+	return newReader(base, overlay, length)
+}
+
 func (or *overlayReader) seek(index int64) (err error) {
 	if or.Overlay.contains(index) {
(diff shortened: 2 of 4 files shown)


ohadlevy:
:+1:  just double checking I understand the usage case, in this case, a pxe boot (using ipxe) will download dynamically the initrd which will end up including the ignition data, and the ignition.config.url= will not be used? will that work with minimized iso usage case too (e.g. coreos.live.rootfs_url option) or that irrelevant as we are always assuming http transport to begin with (ipxe)?

pkg/isoeditor/initramfs.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"io"
+	"os"
+
+	"github.com/openshift/assisted-image-service/pkg/overlay"
+	"github.com/pkg/errors"
+)
+
+func NewInitRamFSStreamReader(irfsPath string, ignitionContent *IgnitionContent) (io.ReadSeeker, error) {
```
omertuc, line 11: Is this function unused on purpose?

pkg/isoeditor/initramfs.go
```diff
@@ -0,0 +1,27 @@
...
+	irfsReader, err := os.Open(irfsPath)
```
carbonin, line 12: This is assuming the initramfs exists locally rather than extracting it from the iso?

Wondering if that fits an "isoeditor" package.

pkg/isoeditor/initramfs.go
carbonin: This is a new public api.
@zaneb isn't using this as a service, but as a library.

pkg/overlay/overlay.go
```diff
@@ -55,6 +48,36 @@ func NewOverlayReader(base io.ReadSeeker, overlay Overlay) (io.ReadSeeker, error
...
+	if err != nil {
+		return nil, err
+	}
+	if overlay.Offset < 0 || overlay.Offset > length {
+		return nil, errors.New("Overlay offset is beyond end of base")
+	}
+	return newReader(base, overlay, length)
+}
+
+func NewAppendReader(base io.ReadSeeker, reader io.ReadSeeker) (io.ReadSeeker, error) {
```
carbonin, line 62: Can't we just use a [MultiReader](https://pkg.go.dev/io@go1.17.3#MultiReader) for this?

pkg/isoeditor/initramfs.go
omertuc: Ack

carbonin (changes requested):
I'm not sure this really needs to be added to the public API of this project.

All the logic you need to reuse (mainly `IgnitionContent.Archive()` IIUC) is already public.
If you remove the changes to overlay in favor of a [MultiReader](https://pkg.go.dev/io@go1.17.3#MultiReader) I think the actual logic added here would be quite small and not particularly complicated.

pkg/isoeditor/initramfs.go
zaneb: Yeah, the plan is to let coreos-installer [do the extraction](https://github.com/openshift/machine-os-images/blob/main/scripts/copy-pxe).
I agree that naming is getting a little awkward This function is very much a mirror of the equivalent ISO one so I feel like they belong in the same package, but I didn't want to go renaming stuff without getting your input first.

pkg/overlay/overlay.go
zaneb: I was hoping so too, but it returns a `Reader` interface instead of `ReadSeeker` because golang :roll_eyes:

carbonin: :man_facepalming: right.

carbonin (approved):
Looks good

Talked a bit about if the initramfs editor stuff really belongs here, but it makes sense based on how @zaneb is using this package and it's not all that much to maintain.

We were also wondering if we have a problem because we're never closing the files we open in these `New*StreamReader` functions, but figure that we can refactor that later if it becomes a problem.

PR #50, opened 2021-12-15
isoeditor: support editing FCOS grub.cfg

vrutkovs:
## Description
Discovery ISO might be FCOS based, so `grub.cfg` needs to be looked up in a different path. This PR checks which of two paths - RHCOS and FCOS - is available and uses it throughout the function

## How was this code tested?
Tested AI flow with FCOS 34 as a base for ISO image

## Assignees
/cc @carbonin
/cc @rollandf 

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -49,6 +49,20 @@ var (
 			"rootfs_url":        "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest/rhcos-live-rootfs.x86_64.img",
 			"version":           "x86_64-latest",
 		},
+		{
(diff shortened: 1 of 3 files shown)


eranco74 (commented):
Looks good.
Please add UT
https://github.com/openshift/assisted-image-service/blob/b910b5cd5617aa374acf2e91b59c82d9516304a8/pkg/isoeditor/rhcos_test.go#L73

carbonin:
+1 to adding unit tests.
Also is there a public download location for fcos that we could add to the integration test images?
https://github.com/openshift/assisted-image-service/blob/725ace298b84e71cfa4be6be364f7483692cc1e0/integration_test/images_test.go#L30-L52

carbonin (approved):

PR #53, opened 2022-01-14
Bug 2034686: Allow for disabling TLS verification for fetching ISOs

CrystalChun:
## Description

[Bugzilla 2034686](https://bugzilla.redhat.com/show_bug.cgi?id=2034686): Allow for disabling TLS verification for fetching ISOs

## How was this code tested?

Manually tested by creating an nginx server secured with a self-signed certificate in the assisted-image-service pod, hosting a file, and updating the assisted-service-config configmap's OS_IMAGES to point to the file url.

## Assignees

/cc @carbonin 
/cc @

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -200,7 +200,7 @@ var _ = BeforeSuite(func() {
 	imageDir, err = ioutil.TempDir("", "imagesTest")
 	Expect(err).To(BeNil())
 
-	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir), imageDir, versions)
+	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir), imageDir, false, versions)
 	Expect(err).NotTo(HaveOccurred())
 
 	err = imageStore.Populate(context.Background())
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ var Options struct {
 	RHCOSVersions         string `envconfig:"RHCOS_VERSIONS"`
 	OSImages              string `envconfig:"OS_IMAGES"`
 	AllowedDomains        string `envconfig:"ALLOWED_DOMAINS"`
+	InsecureSkipVerify    bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
 }
 
 func main() {
@@ -52,7 +53,7 @@ func main() {
 		}
 	}
 
-	is, err := imagestore.NewImageStore(isoeditor.NewEditor(Options.DataDir), Options.DataDir, versions)
+	is, err := imagestore.NewImageStore(isoeditor.NewEditor(Options.DataDir), Options.DataDir, Options.InsecureSkipVerify, versions)
 	if err != nil {
 		log.Fatalf("Failed to create image store: %v\n", err)
 	}
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -2,6 +2,7 @@ package imagestore
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net/http"
@@ -68,9 +69,10 @@ type ImageStore interface {
 }
 
 type rhcosStore struct {
-	versions  []map[string]string
-	isoEditor isoeditor.Editor
-	dataDir   string
+	versions   []map[string]string
+	isoEditor  isoeditor.Editor
+	dataDir    string
+	httpClient *http.Client
 }
 
 const (
@@ -78,14 +80,20 @@ const (
 	ImageTypeMinimal = "minimal-iso"
 )
 
-func NewImageStore(ed isoeditor.Editor, dataDir string, versions []map[string]string) (ImageStore, error) {
+func NewImageStore(ed isoeditor.Editor, dataDir string, insecureSkipVerify bool, versions []map[string]string) (ImageStore, error) {
 	if err := validateVersions(versions); err != nil {
 		return nil, err
 	}
+	transportConfig := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: insecureSkipVerify}, //nolint:gosec // Optionally ignore TLS (G402 error)
+	}
+	httpClient := &http.Client{Transport: transportConfig}
+
 	return &rhcosStore{
-		versions:  versions,
-		isoEditor: ed,
-		dataDir:   dataDir,
+		versions:   versions,
+		isoEditor:  ed,
+		dataDir:    dataDir,
+		httpClient: httpClient,
 	}, nil
 }
 
@@ -115,8 +123,9 @@ func validateVersions(versions []map[string]string) error {
 	return nil
 }
 
-func downloadURLToFile(url string, path string) error {
-	resp, err := http.Get(url)
+func (s *rhcosStore) downloadURLToFile(url string, path string) error {
+	resp, err := s.httpClient.Get(url)
+
 	if err != nil {
 		return err
 	}
@@ -170,7 +179,7 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 				url := imageInfo["url"]
 				log.Infof("Downloading iso from %s to %s", url, fullPath)
 
-				err = downloadURLToFile(url, fullPath)
+				err = s.downloadURLToFile(url, fullPath)
 				if err != nil {
 					return fmt.Errorf("failed to download %s: %v", url, err)
 				}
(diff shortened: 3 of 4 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -109,9 +111,24 @@ func validateVersions(versions []map[string]string) error {
 }
 
 func downloadURLToFile(url string, path string) error {
-	resp, err := http.Get(url)
+	var resp *http.Response
+	var err error
+
+	resp, err = http.Get(url)
```
romfreiman, line 117: Isn't there just a flag not to verify tls?

romfreiman: I'm pretty sure we do it in the agent alread'

carbonin: Yeah, that's what the block below is doing.
@CrystalChun we can probably just do the one call rather than waiting for an error and then doing the unverified one.

Also it might be better to create this client in `NewImageStore` and use it here rather than recreating it for each request.

CrystalChun: Yep, the flag in the [agent](https://github.com/openshift/assisted-installer-agent/blob/42d49b73098ae12d1879768f01c390fa952d1cae/src/session/inventory_session.go#L74) is set the same way. 

Good idea, I've moved the client creation to the `NewImageStore` function.

pkg/imagestore/imagestore.go
```diff
@@ -53,6 +54,8 @@ var DefaultVersions = []map[string]string{
 	},
 }
 
+var httpClient *http.Client
```
filanov, line 57: If `downloadURLToFile`  is used only by `rhcosStore` does it make sense move this variable and function under `rhcosStore` somehow global doesn't feels right here.

carbonin: :+1: yes, making this part of the `rhcosStore` struct is what I had in mind

pkg/imagestore/imagestore.go
```diff
@@ -75,6 +78,12 @@ func NewImageStore(ed isoeditor.Editor, dataDir string, versions []map[string]st
 	if err := validateVersions(versions); err != nil {
 		return nil, err
 	}
+	// Create an HTTP Client that always skips TLS
+	transportConfig := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec // Solution is to ignore TLS (G402 error)
```
flaper87, line 83: Not sure if this was discussed elsewhere but, should this be configured? Not all deployments may want to run insecure TLS traffic. Am I missing something?

carbonin: Yeah, I suppose this should be configurable.
For example we probably don't ever want this behavior in the cloud.

pkg/imagestore/imagestore.go
CrystalChun: Thanks for the clarification! I'll go ahead and move it under the `rhcosStore` struct

pkg/imagestore/imagestore.go
CrystalChun: Good point! Will make this configurable

romfreiman: What will be the default though? Currently we dont have a way to provide a certificate - so the user has not choice but running it in this mode. In the future, if we decide to add a certificate, it will make sense to have it as a configuration option, until then, u either have it insecure or serve it on http, so not sure how configuration will help here.

romfreiman: @carbonin why do we care about the cloud - we deploy it, no? how would it point to different values then those that we provide...

carbonin: Right, and if the certs don't verify on `mirror.openshift.com` then I feel like something is wrong.
We still want MITM protection if we can get it.

People who need this feature are deploying the mirror themselves and know that we are not going to be able to verify that cert. We could make this (insecure) the default for the operator deploy or pass the configuration through the operator in some other way.

Another secure option would be to also have the file sha sum specified in os images. Then we can verify the content without being concerned that it isn't what we expected. These are provided on mirror.openshift.com (e.g. https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/latest/latest/sha256sum.txt), but this is a bit of a larger change.

CrystalChun: Is it best to verify certs by default?

main.go
```diff
@@ -27,6 +27,7 @@ var Options struct {
 	RHCOSVersions         string `envconfig:"RHCOS_VERSIONS"`
 	OSImages              string `envconfig:"OS_IMAGES"`
 	AllowedDomains        string `envconfig:"ALLOWED_DOMAINS"`
+	InsecureSkipVerify    string `envconfig:"INSECURE_SKIP_VERIFY"`
```
carbonin, line 30: This can be a bool with default set to `false`.
Here's an example from assisted-service: https://github.com/openshift/assisted-service/blob/80170758e2463f3b0f7e1c0256f1ebf7be57d0ab/cmd/main.go#L146

main.go
```diff
@@ -52,7 +53,17 @@ func main() {
...
-	is, err := imagestore.NewImageStore(isoeditor.NewEditor(Options.DataDir), Options.DataDir, versions)
+	var insecureSkipVerify bool
+	insecureSkipVerifyOption := Options.InsecureSkipVerify
+	if insecureSkipVerifyOption == "" || insecureSkipVerifyOption == "false" {
+		insecureSkipVerify = false
+	} else if insecureSkipVerifyOption == "true" {
+		insecureSkipVerify = true
+	} else {
+		log.Fatalf("Unrecognizable option for InsecureSkipVerify.")
+	}
```
carbonin, line 64: This can all be removed if the option is a bool.

integration_test/images_test.go
```diff
@@ -186,7 +186,7 @@ var _ = BeforeSuite(func() {
 	imageDir, err = ioutil.TempDir("", "imagesTest")
 	Expect(err).To(BeNil())
 
-	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir), imageDir, versions)
+	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir), imageDir, true, versions)
```
carbonin, line 189: I'd expect this to false, right?
We would like to verify the official mirror I'd think.

pkg/imagestore/imagestore.go
```diff
@@ -108,8 +116,9 @@ func validateVersions(versions []map[string]string) error {
 	return nil
 }
 
-func downloadURLToFile(url string, path string) error {
-	resp, err := http.Get(url)
+func downloadURLToFile(httpClient *http.Client, url string, path string) error {
```
carbonin, line 119: Is there any reason not to make this a method on `rhcosStore` rather than passing in the client?

pkg/imagestore/imagestore_test.go
```diff
@@ -76,7 +76,7 @@ var _ = Context("with a data directory configured", func() {
 					),
 				)
 				version["url"] = ts.URL() + "/some.iso"
-				is, err := NewImageStore(mockEditor, dataDir, []map[string]string{version})
+				is, err := NewImageStore(mockEditor, dataDir, true, []map[string]string{version})
```
carbonin, line 79: These were working before so I'd rather keep these as the default (`false`) whenever we can.
Even in tests let's make setting the insecure options the exception.

integration_test/images_test.go
filanov: usually it's clearer to set a type for a boolean value, this way it's clear what is it used for, for example https://github.com/openshift/assisted-service/blob/80170758e2463f3b0f7e1c0256f1ebf7be57d0ab/internal/common/db.go#L119-L124

pkg/imagestore/imagestore_test.go
CrystalChun: Gotcha, changed it back to `false`.

integration_test/images_test.go
CrystalChun: @carbonin Yes! Thanks for catching that, I've changed it to `false` to verify the official mirror.

Thanks @filanov! I'll keep that in mind for future variables.

main.go
CrystalChun: Removed it, thank you!

main.go
CrystalChun: Modified, thank you for the example!

flaper87:
/retitle Bug 2034686: Allow for disabling TLS verification for fetching ISOs

flaper87 (commented):
Looks good! I'd recommend updating the description for the commit and PR to match what the patch does. It's moved from ignoring errors to allowing to disable TLS verification.

Thanks for the work here :pray:

CrystalChun:
Thank you @flaper87! Just updated both the commit and the PR description :)

carbonin (approved):

PR #55, opened 2022-01-16
MGMT-8925: allow more size in requested PV to store all images

osherdp:
## Description

Seems like with the addition of 4.10.0-fc.0 image we also got  to the limit storage size of 10GB.
Trying to increase it to 15GB for now, so we can have a fixed integration environment:

## How was this code tested?

Will be tested after automatic deployment of the changes.

## Assignees

/cc @filanov 
/cc @carbonin 

## Links

https://coreos.slack.com/archives/CCRND57FW/p1642342769294900

## Checklist


filanov (approved):

romfreiman:
@osherdp should we update the docs?

osherdp:
> @osherdp should we update the docs?

Which ones?

romfreiman:
official?

filanov:
Are there any changes in the operator? 
If we are  going to support multiple versions in 2.5 it may be affected as well

osherdp:
At least in here it tells you to request ``20GB`` for images:
https://github.com/openshift/assisted-service/blob/eb983e2b92936a03190c81077aa04e43945157e1/docs/operator.md#deploying-the-operand

If we want to handle multiple versions, then we need to tell the user how to calculate the needed space (or we should calculate it and update the doc with each operator release)
That's because it's probably not just the raw size of the images, but some more work space is needed at start.

filanov:
I think that adding a documentation to how to calculate the size is a good ides. 
Another option is resolve it in the code, we will need to check how long it will take to extract the minimal ISO, if it's fast because it is working on a local filesystem then we may do it serial then we can reduce the requirement.

filanov:
from a test that i run it looks like creating minimal iso is really fast so maybe (if i'm right about where is the size is increasing) we can do this in serial?
```
time="2022-01-17T10:43:59+02:00" level=info msg="Downloading iso from https://mirror.openshift.com/pub/openshift-v4/arm64/dependencies/rhcos/pre-release/latest/rhcos-live.aarch64.iso to /tmp/imagesTest3588835598/rhcos-full-iso-pre-release-arm-latest-arm64.iso"
time="2022-01-17T10:45:52+02:00" level=info msg="Finished downloading for pre-release-arm64 (arm-latest)"
time="2022-01-17T10:45:52+02:00" level=info msg="full iso hanling took%!(EXTRA time.Duration=1m53.03646076s)"
time="2022-01-17T10:45:52+02:00" level=info msg="Creating minimal iso for pre-release-arm-latest-arm64"
time="2022-01-17T10:45:53+02:00" level=info msg="Finished creating minimal iso for pre-release-arm64 (arm-latest)"
time="2022-01-17T10:45:53+02:00" level=info msg="minimal iso hanlding took%!(EXTRA time.Duration=544.825967ms)"
```

osherdp:
@filanov do you want to create the task to do that serially?
or alternatively schedule the meeting with me and @carbonin to discuss the alternatives

carbonin:
The important part to do in parallel is downloading the full iso.
If we want to do the minimal iso creation serially after they're all downloaded then that shouldn't be too much of a change.

That would take the storage down from twice the space we need to just having one iso's worth of extra space, right?

osherdp:
sounds like it

osherdp:
sounds like it

filanov:
opened https://issues.redhat.com/browse/MGMT-9019

PR #61, opened 2022-02-11
Use fields from http.DefaultTransport in imagestore

carbonin:
## Description

http.DefaultTransport has some useful properties set that are not
present in a zero-value transport struct.

Specifically it sets up a function for fetching proxy environment
variables.

When we moved from using the default transport to creating our own to
solve https://bugzilla.redhat.com/show_bug.cgi?id=2034686 we also
removed the proxy environment variable handling as a side-effect.

## How was this code tested?

This was verified to fix the proxy issue in @chadcrum 's test environment.

## Assignees

/cc @CrystalChun 
/cc @danielerez 

## Links

https://bugzilla.redhat.com/show_bug.cgi?id=2038903

## Checklist


CrystalChun (approved):
/lgtm

PR #62, opened 2022-02-24
MGMT-9332: Add endpoint for rootfs and kernel

carbonin:
## Description

This PR adds a new endpoint to the image service
which will serve the rootfs or the kernel directly
from an ISO.

The new endpoint is of the form:
```
/boot-artifacts/{artifact-name}?{params}
```

For example:
```
/boot-artifacts/rootfs?version=4.9&arch=x86_64
/boot-artifacts/kernel?version=4.10&arch=arm64
```

## How was this code tested?

Ran the image service locally and downloaded the rootfs and kernel using the new endpoints.
Ensured that they matched exactly the same files extracted from the iso using the `iso-read` command line tool.

## Assignees

/cc @mhrivnak 
/cc @danielerez 

## Links

Enhancement doc: https://github.com/openshift/assisted-service/pull/3389
JIRA: https://issues.redhat.com/browse/MGMT-9332

## Checklist

### DIFF
--- a/internal/handlers/boot_artifacts.go
+++ b/internal/handlers/boot_artifacts.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/openshift/assisted-image-service/pkg/imagestore"
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
@@ -39,9 +40,10 @@ func parseArtifact(path string) (string, error) {
 }
 
 func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.Method != "GET" {
-		log.Error("Only GET method is supported with this endpoint.")
-		http.NotFound(w, r)
+	if r.Method != http.MethodGet && r.Method != http.MethodHead {
+		log.Error("Only GET and HEAD methods are supported with this endpoint.")
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		w.Header().Set("Allow", strings.Join([]string{http.MethodGet, http.MethodHead}, ", "))
 		return
 	}
 
--- a/internal/handlers/boot_artifacts.go
+++ b/internal/handlers/boot_artifacts.go
@@ -1,0 +1,97 @@
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"regexp"
+	"time"
+
+	"github.com/openshift/assisted-image-service/pkg/imagestore"
+	"github.com/openshift/assisted-image-service/pkg/isoeditor"
+	log "github.com/sirupsen/logrus"
+)
+
+type BootArtifactsHandler struct {
+	ImageStore imagestore.ImageStore
+}
+
+var _ http.Handler = &BootArtifactsHandler{}
+
+var bootpathRegexp = regexp.MustCompile(`^/boot-artifacts/(.+)`)
+
+func parseArtifact(path string) (string, error) {
+	match := bootpathRegexp.FindStringSubmatch(path)
+	if len(match) < 1 {
+		return "", fmt.Errorf("malformed download path: %s", path)
+	}
+
+	var artifact string
+	switch match[1] {
+	case "rootfs":
+		artifact = "rootfs.img"
+	case "kernel":
+		artifact = "vmlinuz"
+	default:
+		return "", fmt.Errorf("malformed download path: %s", path)
+	}
+	return artifact, nil
+}
+
+func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "GET" {
+		log.Error("Only GET method is supported with this endpoint.")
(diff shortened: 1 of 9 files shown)


carbonin:
Can someone else review this since @CrystalChun and I effectively paired on it and @mhrivnak is out?

Maybe @danielerez, @avishayt or @filanov (or maybe one of you could suggest someone that could look it over?)

internal/handlers/boot_artifacts.go
```diff
@@ -0,0 +1,99 @@
...
+		artifact = "rootfs.img"
+	case "kernel":
+		artifact = "vmlinuz"
+	default:
+		return "", fmt.Errorf("malformed download path: %s", path)
+	}
+	return artifact, nil
+}
+
+func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
romfreiman, line 41: metrics? events?

carbonin: Yeah, we should add this to the metrics. Not sure what you mean by events though.

carbonin:
@romfreiman 's question about metrics also reminded me of the issues we had with CORS.
We don't expect anyone to be downloading these files directly from the UI, right?

internal/handlers/boot_artifacts.go
romfreiman: Hm. I think we have events on generating an iso. So is it applicable here as well?

romfreiman:
I assume we gonna present the script in he ui at some point, won't we? And allow to download it. 
About rootfs/kernel, why not? Similar to what we do with the iso

carbonin:
> I assume we gonna present the script in he ui at some point, won't we? And allow to download it.

The script is going to be served by assisted-service, so CORS is already handled there for all endpoints I think.

> About rootfs/kernel, why not?

:man_shrugging: I'll add CORS headers for this also then.

carbonin:
Marked a WIP to work on metrics and CORS.

internal/handlers/boot_artifacts.go
```diff
@@ -0,0 +1,99 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		_, err = w.Write([]byte(err.Error()))
+		if err != nil {
+			log.Errorf("Failed to write response: %v\n", err)
+		}
+		return
+	}
+
+	isoFileName := b.ImageStore.PathForParams(imagestore.ImageTypeFull, version, arch)
+	fileReader, err := isoeditor.GetFileFromISO(isoFileName, fmt.Sprintf("/images/pxeboot/%s", artifact))
```
avishayt, line 65: "/images/pxeboot" already appears several times in the repo, consider defining it as a const

internal/handlers/boot_artifacts.go
```diff
@@ -0,0 +1,99 @@
...
+	defer fileReader.Close()
+
+	fileInfo, err := os.Stat(isoFileName)
+	if err != nil {
+		log.Errorf("Error reading file info for %s", isoFileName)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", artifact))
```
avishayt, line 80: What about the Content-Length header?

avishayt (commented):
Overall looks great, just a couple comments

internal/handlers/boot_artifacts.go
```diff
@@ -0,0 +1,99 @@
...
+		log.Errorf("Error creating file reader stream: %v\n", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	defer fileReader.Close()
+
+	fileInfo, err := os.Stat(isoFileName)
+	if err != nil {
+		log.Errorf("Error reading file info for %s", isoFileName)
+		w.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
carbonin: In this case just the error code, but a failure to stat a file isn't really something a user can do anything about.

We can respond with a message if you think it will be helpful though.

internal/handlers/boot_artifacts.go
carbonin: Content-Length is handled by `ServeContent`.
But this reminds me that we should also be allowing `HEAD` in the check at the top of this function.

internal/handlers/boot_artifacts.go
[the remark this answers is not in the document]
filanov: it's more helpful for us i think, when someone will complain it could save us some time debugging no?

internal/handlers/boot_artifacts.go
carbonin: This is the only occurrence that isn't in tests.

Specifically most uses are mocking or asserting on grub config. I'm not sure it's worth doing string interpolation for all of those.

internal/handlers/boot_artifacts.go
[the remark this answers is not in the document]
carbonin: There's already a helper for this https://pkg.go.dev/net/http@go1.17.7#Error :man_facepalming:

PR #63, opened 2022-02-28
Fix overlay reader for base readers that return EOF early

carbonin:
## Description

It's valid for a reader to return EOF at the same time as the last
bytes. If we have such a reader as the base for an overlay we return EOF
when the base reader returns EOF even if we haven't printed the overlay
yet.

This is specifically a problem for the append reader when we're always
getting to the end of the base before starting to read the overlay.

## How was this code tested?

Unit and integration tests cover this

## Assignees

/cc @zaneb 
/cc @filanov 

## Links

https://issues.redhat.com/browse/MGMT-9333

## Checklist

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -143,7 +143,7 @@ func (or *overlayReader) Read(p []byte) (int, error) {
 
 	seekErr := or.seek(or.readIndex + int64(bytesRead))
 
-	if seekErr == nil || (readErr != nil && readErr != io.EOF) {
+	if readErr != nil && readErr != io.EOF {
 		return bytesRead, readErr
 	}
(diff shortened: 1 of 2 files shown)


carbonin:
/hold

Going to put a hold on it until I test with my ignition+initrd branch

zaneb:
/approve
We don't _appear_ to have hit this in image-customization-controller, so I'd be interested to know what set of circumstances triggers it in case we need to backport.

carbonin:
I think we're reading exactly the number of bytes we need to the end of the base file and _most_ reader implementations don't return EOF in that case.

As for higher-level symptoms, I was seeing the download cut off before the appended portion of the reader.

filanov (approved):

PR #64, opened 2022-03-01
MGMT-9333 Add ipxe-initrd endpoint for customized initrd

carbonin:
## Description

This PR creates a new endpoint in the image service which will serve an initrd which includes the ignition for a particular infra-env.

The new endpoint is of the form:

```
/images/{image-id}/pxe-initrd?{params}
```

For example:

```
/images/dbe345ff-a89e-4c61-878f-d50066d1af3a/pxe-initrd?version=4.9&arch=x86_64
```

The restructuring of the image handler is worth calling out specifically.
The go standard library http server implementation works only on path prefixes.
This means that to implement a sub-path of a particular resource (such as this one) we need an intermediate handler that handles parsing the path and calling the correct sub-handler.

This PR turns the image handler into this "base" handler that will route to either the iso or initrd download handler.

## How was this code tested?

~~WIP: Still looking for a way to verify this manually but parsing out the concatenated initrd format seems non-trivial.~~
It was non-trivial but this is what I did:

- Configure this branch of the image service to run against the integration saas assisted service
- Create a cluster in integration UI and fetch the infra-env ID
- Download the pxe-initrd from the local image service
- Use [`binwalk`](https://www.kali.org/tools/binwalk/) on the resulting file to find the last gzipped archive location
```
$ binwalk 02bbb814-eafa-497b-811e-86931d697d33-initrd.img

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
0             0x0             ASCII cpio archive (SVR4 with no CRC), file name: ".", file name length: "0x00000002", file size: "0x00000000"
112           0x70            ASCII cpio archive (SVR4 with no CRC), file name: "early_cpio", file name length: "0x0000000B", file size: "0x00000002"
240           0xF0            ASCII cpio archive (SVR4 with no CRC), file name: "kernel", file name length: "0x00000007", file size: "0x00000000"
360           0x168           ASCII cpio archive (SVR4 with no CRC), file name: "kernel/x86", file name length: "0x0000000B", file size: "0x00000000"
484           0x1E4           ASCII cpio archive (SVR4 with no CRC), file name: "kernel/x86/microcode", file name length: "0x00000015", file size: "0x00000000"
616           0x268           ASCII cpio archive (SVR4 with no CRC), file name: "kernel/x86/microcode/AuthenticAMD.bin", file name length: "0x00000026", file size: "0x00007752"
31312         0x7A50          ASCII cpio archive (SVR4 with no CRC), file name: "kernel/x86/microcode/GenuineIntel.bin", file name length: "0x00000026", file size: "0x005B9800"
6034148       0x5C12E4        ASCII cpio archive (SVR4 with no CRC), file name: "TRAILER!!!", file name length: "0x0000000B", file size: "0x00000000"
6034432       0x5C1400        gzip compressed data, maximum compression, from Unix, last modified: 1970-01-01 00:00:00 (null date)
14185156      0xD872C4        MySQL ISAM index file Version 3
20533552      0x1395130       gzip compressed data, from NTFS filesystem (NT), last modified: 1970-01-01 00:00:00 (null date)
20547306      0x13986EA       gzip compressed data, from NTFS filesystem (NT), last modified: 1970-01-01 00:00:00 (null date)
20551203      0x1399623       gzip compressed data, from NTFS filesystem (NT), last modified: 1970-01-01 00:00:00 (null date)
...
87837436      0x53C4AFC       gzip compressed data, has original file name: "dracut-random.cpio", from Unix, last modified: 2019-12-10 14:51:33
87837607      0x53C4BA7       gzip compressed data, maximum compression, from Unix, last modified: 1970-01-01 00:00:00 (null date)
87854752      0x53C8EA0       gzip compressed data, last modified: 1970-01-01 00:00:00 (null date)
```
- Extract the ignition and verify it is correct:
```
$ dd if=02bbb814-eafa-497b-811e-86931d697d33-initrd.img of=ignition.img.gz skip=87854752 bs=1
4144+0 records in
4144+0 records out
4144 bytes (4.1 kB, 4.0 KiB) copied, 0.0109975 s, 377 kB/s
$ file ignition.img.gz 
ignition.img.gz: gzip compressed data, original size modulo 2^32 10248
$ gunzip ignition.img.gz 
$ cpio -idmv --no-absolute-filenames < ignition.img 
config.ign
21 blocks
$ cat config.ign | jq
{
  "ignition": {
    "version": "3.1.0"
  },
  "passwd": {
    "users": []
  },
  "systemd": {
    "units": [
      {
        "name": "agent.service",
        "enabled": true,
...
```

## Assignees

/cc @CrystalChun 
/cc @filanov 
/cc @avishayt 

## Links

~~Depends on https://github.com/openshift/assisted-image-service/pull/62~~ Merged
Resolves https://issues.redhat.com/browse/MGMT-9333

## Checklist


internal/handlers/assisted_service_client.go
```diff
@@ -0,0 +1,143 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type AssistedServiceClient struct {
+	assistedServiceScheme string
+	assistedServiceHost   string
+	client                *http.Client
+}
+
+const fileRouteFormat = "/api/assisted-install/v2/infra-envs/%s/downloads/files"
```
filanov, line 22: Isn't it better to import the existing client? similar to the installer and the agent?

carbonin: If we published the client separately then sure, but I don't want to have assisted-service as a dependency for two endpoints.

filanov: my pr was pushed, we can probably add the client as a sub module as well it should be easy now.
Will try to work on that. 
I will check what controller and installer are using

carbonin: Yeah I think using the client would be the correct direction long term, but since this PR is just moving these functions around rather than adding anything new I'd rather leave changing the client around for a follow-up effort.

Specifically I'm concerned that the way we push auth from the image service request through to the assisted request may be difficult to translate to the generated client. I might be wrong, but we'd need to investigate this.

filanov: good point about auth, by default it is set on creation.

internal/handlers/assisted_service_client.go
```diff
@@ -0,0 +1,143 @@
...
+
+func NewAssistedServiceClient(assistedServiceScheme, assistedServiceHost, caCertFile string) *AssistedServiceClient {
+	client := &http.Client{}
+	if caCertFile != "" {
+		caCert, err := ioutil.ReadFile(caCertFile)
+		if err != nil {
+			log.Fatalf("Error opening cert file %s, %s", caCertFile, err)
```
filanov, line 29: fatal will crash the service usually it recommended to return and error and handle it in main.

carbonin: This is just moved from https://github.com/openshift/assisted-image-service/pull/64/files#diff-5e672f28c4fa21931ec9a0d5438b1cf4deb15400b449463d9d8e77f879a6b0a2L53-L69

I think we should crash if we can't open this file, but if you'd like me to move the fatal log to main, I can do that.

filanov: Not for this PR but lets open a ticket.

filanov (approved):

internal/handlers/assisted_service_client.go
carbonin: :+1: opened https://issues.redhat.com/browse/MGMT-9538

PR #66, opened 2022-03-25
Return a 503 from images and boot-artifacts before downloads finish

carbonin:
## Description

This commit extends the readiness handler to provide a middleware that
will return 503s until we mark the handler ready. It will then call the
underlying handler.

This new middleware is applied to both the images and boot-artifacts
endpoints so it's more clear to a user that the failure it temporary
rather than some issue with the deployment or configuration.

## How was this code tested?

Deployed the service locally and tested responses from both the images and boot-artifacts endpoints before and after the template images were finished downloading.

## Assignees

/cc @CrystalChun 
/cc @filanov 

## Links

https://issues.redhat.com/browse/MGMT-9395

## Checklist

### DIFF
--- a/internal/handlers/readiness.go
+++ b/internal/handlers/readiness.go
@@ -17,11 +17,22 @@ func NewReadinessHandler() *ReadinessHandler {
 }
 
 func (a *ReadinessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ok := func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }
+	a.runIfReady(http.HandlerFunc(ok), w, r)
+}
+
+func (a *ReadinessHandler) WithMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		a.runIfReady(next, w, r)
+	})
+}
+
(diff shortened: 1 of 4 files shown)


internal/handlers/readiness.go
```diff
@@ -17,11 +17,22 @@ func NewReadinessHandler() *ReadinessHandler {
...
+func (a *ReadinessHandler) runIfReady(next http.Handler, w http.ResponseWriter, r *http.Request) {
 	if !a.isEnabled {
 		w.WriteHeader(http.StatusServiceUnavailable)
 		return
 	}
-	w.WriteHeader(http.StatusOK)
+	next.ServeHTTP(w, r)
 }
 
 func (a *ReadinessHandler) Enable() {
```
CrystalChun, line 38: Small question is do we want to keep the log in this function? If we do, is it a good idea to log which API is enabled since this middleware is now used by multiple handlers?

Looks good otherwise! I'll wait until @filanov reviews

filanov: if i understand correctly it mean that all the apis are enabled and now the service ready to serve requests

filanov (approved):

internal/handlers/readiness.go
carbonin: Correct. The same `Enable` call applies to all the handlers we wrap in the middleware.

CrystalChun: Gotcha, thanks for the explanation!

PR #67, opened 2022-03-29
[release-ocm-2.4] Bug 2069806: Use fields from http.DefaultTransport in imagestore

carbonin:
## Description

http.DefaultTransport has some useful properties set that are not
present in a zero-value transport struct.

Specifically it sets up a function for fetching proxy environment
variables.

When we moved from using the default transport to creating our own to
solve https://bugzilla.redhat.com/show_bug.cgi?id=2034686 we also
removed the proxy environment variable handling as a side-effect.

This is a backport to ocm-2.4 of https://github.com/openshift/assisted-image-service/pull/61

## Assignees

/cc @CrystalChun 
/cc @filanov 

## Links

https://bugzilla.redhat.com/show_bug.cgi?id=2069806

## Checklist


CrystalChun (approved):
/lgtm

PR #68, opened 2022-04-11
MGMT-9538: Only log.Fatal from main in the image service

mresvanis:
## Description

This PR ensures that we use `log.Fatalf` only in the `main` function and not from internal or other package functions.

## Assignees

/cc @carbonin 

## Checklist


carbonin (commented):
Don't ignore errors in tests.
Otherwise looks good.

mresvanis:
Very good point, updated accordingly.

carbonin (approved):

PR #71, opened 2022-04-25
Bug 2078531: add an option expose all endpoints on plain HTTP port

vrutkovs:
## Description
IPXE flow requires artifacts to be available via HTTP. This change would ensure that if `LISTEN_HTTP_PORT` env var is set and HTTPS cert/key are set, the same endpoints are exposed on plain HTTP port as well

TODO:
* [x] Add a wrapper around http.Server and add unit tests

## How was this code tested?
local deployment

## Assignees
/cc @carbonin 
/cc @CrystalChun 
/cc @filanov 

## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ var Options struct {
 	HTTPSCertFile         string `envconfig:"HTTPS_CERT_FILE"`
 	HTTPSCAFile           string `envconfig:"HTTPS_CA_FILE"`
 	ListenPort            string `envconfig:"LISTEN_PORT" default:"8080"`
+	HTTPListenPort        string `envconfig:"HTTP_LISTEN_PORT"`
 	MaxConcurrentRequests int64  `envconfig:"MAX_CONCURRENT_REQUESTS" default:"400"`
 	RHCOSVersions         string `envconfig:"RHCOS_VERSIONS"`
 	OSImages              string `envconfig:"OS_IMAGES"`
@@ -106,11 +107,23 @@ func main() {
 	http.Handle("/live", handlers.NewLivenessHandler())
 	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
 
-	log.Info("Starting http handler...")
-	address := fmt.Sprintf(":%s", Options.ListenPort)
+	httpsListen := func(port string) {
+		log.Infof("Starting https handler on %s...", port)
+		log.Fatal(http.ListenAndServeTLS(fmt.Sprintf(":%s", port), Options.HTTPSCertFile, Options.HTTPSKeyFile, nil))
+	}
+
+	httpListen := func(port string) {
+		log.Infof("Starting http handler on %s...", port)
+		log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+	}
+
 	if Options.HTTPSKeyFile != "" && Options.HTTPSCertFile != "" {
-		log.Fatal(http.ListenAndServeTLS(address, Options.HTTPSCertFile, Options.HTTPSKeyFile, nil))
+		go httpsListen(Options.ListenPort)
+		if Options.HTTPListenPort != "" {
+			go httpListen(Options.HTTPListenPort)
+		}
+		select {}
 	} else {
-		log.Fatal(http.ListenAndServe(address, nil))
+		httpListen(Options.ListenPort)
 	}
 }
--- a/main.go
+++ b/main.go
@@ -5,6 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/kelseyhightower/envconfig"
 	"github.com/openshift/assisted-image-service/internal/handlers"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -26,6 +26,7 @@ var Options struct {
 	HTTPSCertFile         string `envconfig:"HTTPS_CERT_FILE"`
 	HTTPSCAFile           string `envconfig:"HTTPS_CA_FILE"`
 	ListenPort            string `envconfig:"LISTEN_PORT" default:"8080"`
+	PlainHTTPListenPort   string `envconfig:"LISTEN_HTTP_PORT"`
```
carbonin, line 29: I'd make the variable and env var match.
So maybe `HTTPListenPort` and `HTTP_LISTEN_PORT`

main.go
```diff
@@ -106,11 +107,24 @@ func main() {
...
+			}()
+			select {}
+		} else {
+			log.Infof("Starting https handler on %s...", Options.ListenPort)
+			log.Fatal(http.ListenAndServeTLS(fmt.Sprintf(":%s", Options.ListenPort), Options.HTTPSCertFile, Options.HTTPSKeyFile, nil))
+		}
 	} else {
-		log.Fatal(http.ListenAndServe(address, nil))
+		log.Infof("Starting http handler on %s...", Options.ListenPort)
+		log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", Options.ListenPort), nil))
```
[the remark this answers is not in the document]
vrutkovs: Reworkd this:
* define `httpListen` and `httpsListen` so that these can be reused (`httpsListen` is used once, but its easier to keep it uniform)
* in https mode always run `httpsListen` as a goroutine. If `HTTPListenPort` is set we start http listerner as goroutine - and wait for it to be completed in both cases

main.go
```diff
@@ -106,11 +107,23 @@ func main() {
...
+		log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
+	}
+
 	if Options.HTTPSKeyFile != "" && Options.HTTPSCertFile != "" {
-		log.Fatal(http.ListenAndServeTLS(address, Options.HTTPSCertFile, Options.HTTPSKeyFile, nil))
+		go httpsListen(Options.ListenPort)
+		if Options.HTTPListenPort != "" {
+			go httpListen(Options.HTTPListenPort)
+		}
+		select {}
```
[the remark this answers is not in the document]
vrutkovs: Right, this would not wait for goroutines to complete.

I don't think that affects anything, but lets keep it on hold until we confirm this doesn't affect any functionality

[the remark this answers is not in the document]
CrystalChun: Would it also be relevant to take into account if one server errors out? What would happen then?

filanov:
How it will be exposed with the operator? 
Not sure that i understand the end to end flow with this. 
The service is the one that sign the URLs, just want to know what is the plan

vrutkovs:
We had a quick chat with Nick on this, we'll need:
* -image-service to serve content on HTTPS and HTTP (this PR)
* -service change to serve ipxe script via HTTP
* -service change to generate plain http links
* operator change to expose http ports in container/service and create a plain http route when ipxe setting is enabled

We can discuss this in details on arch meeting

filanov:
Sounds good thanks. are there any changes required in auth of the service?

main.go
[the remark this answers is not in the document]
vrutkovs: >if one server errors out

both goroutines will be terminated (ungracefully atm)

carbonin:
> are there any changes required in auth of the service?

Nope, just http vs https

main.go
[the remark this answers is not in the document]
CrystalChun: Maybe we could use a [channel](https://gobyexample.com/channels) to listen for errors and terminate the servers if an error is received? Something like this?
```
         errs := make(chan error)
         httpListen := func(port string) {
		log.Infof("Starting http handler on %s...", port)
		if err := http.ListenAndServe(fmt.Sprintf(":%s", port), nil); err != nil {
                     errs <- err
                 }
         }
...
         select {
         case err := <-errs:
             log.Printf("Could not start serving service due to (error: %s)", err)
             // shut down servers
         }
```

filanov:
Do you plan to add tests?

vrutkovs:
I don't see a nice place for `main.go` tests here - any suggestions? Not sure if `boot_artifacts_test.go` and `initrd_test.go` are the right place for this.

main.go
```diff
@@ -106,11 +110,53 @@ func main() {
...
+		}
+		go httpListen()
+	}
+	<-stop
+	if err := httpsServer.Shutdown(context.TODO()); err != nil {
+		log.Fatalf("HTTPS shutdown failed: %v", err)
+	} else {
+		log.Info("HTTPS server terminated gracefully")
+	}
+	if err := httpServer.Shutdown(context.TODO()); err != nil {
```
carbonin, line 157: These two calls should check for nil I think

main.go
```diff
@@ -106,11 +110,53 @@ func main() {
...
+			}
+			go httpListen()
+		}
+	} else {
+		httpServer = http.Server{
+			Addr: fmt.Sprintf(":%s", Options.ListenPort),
+		}
+		go httpListen()
+	}
+	<-stop
```
carbonin, line 151: Should we have different behavior for TERM vs INT?
`http.Server` has `Close()` and `Shutdown()` for non-graceful/graceful respectively.
Do we think there's a reason to differentiate between the cases?

main.go
vrutkovs: `Shutdown` returning non-nil is an error, I think we want to have this logged at least

main.go
vrutkovs: Lets attempt `Close` when graceful didn't work (or took too long, but that might be too complicated)

main.go
carbonin: No, I mean `httpServer` could be nil. This will segfault

vrutkovs: Hmm, it appears `http.Server` cannot be nil and it won't crash on `Shutdown` when not running - https://go.dev/play/p/Hz-hDxEZlOf

carbonin: Perfect. Just went to look it up and saw:
> The zero value for Server is a valid configuration

ref: https://pkg.go.dev/net/http@go1.18.1#Server

:tada: 

>There is no branch where httpServer could be nil 

Not that it matters now, but it would have been unset in the case where we set the https config, but not http port, right?
Even still it's clear to me now that `http.Server` is not a pointer value so would never be nil for that reason.

main.go
carbonin: I suppose we could do a shutdown with a context, but yeah kube will just kill the container at some point so it may not be worth worrying about.

carbonin:
This looks good to me.

@filanov did you want @avishayt to take a look?

PR #72, opened 2022-04-26
Set Last-Modified time for iso and initrd downloads

carbonin:
## Description
This time is based on the last modified time sent to the service when
querying for the ignition from assisted-service.

This will be used by the ironic cache to determine when it needs to
re-download an image. Before this change the image would be
re-downloaded every time a host booted, but now the cache will start
acting like a cache.

## How was this code tested?

Deployed assisted service and this image service branch locally and observed `Last-Modified` header on iso downloads matching the InfraEnv updated_at time.

## Assignees

/cc @danielerez 
/cc @flaper87 
/cc @CrystalChun 

## Links

https://issues.redhat.com/browse/MGMT-7274

## Checklist

### DIFF
--- a/internal/handlers/assisted_service_client.go
+++ b/internal/handlers/assisted_service_client.go
@@ -91,9 +91,9 @@ func (c *AssistedServiceClient) ramdiskContent(imageServiceRequest *http.Request
 // ignitionContent returns the ignition data on success and the error and the corresponding http status code
 // The code is also returned to ensure issues with authentication from the assisted service request are communicated back to the image service user
 // The returned code should only be used if an error is also returned
-func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string) (*isoeditor.IgnitionContent, int, error) {
+func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string) (*isoeditor.IgnitionContent, string, int, error) {
 	if c.assistedServiceHost == "" {
-		return nil, 0, nil
+		return nil, "", 0, nil
 	}
 
 	u := url.URL{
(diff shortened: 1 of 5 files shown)


internal/handlers/assisted_service_client.go
```diff
@@ -91,9 +91,9 @@ func (c *AssistedServiceClient) ramdiskContent(imageServiceRequest *http.Request
 // ignitionContent returns the ignition data on success and the error and the corresponding http status code
 // The code is also returned to ensure issues with authentication from the assisted service request are communicated back to the image service user
 // The returned code should only be used if an error is also returned
-func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string) (*isoeditor.IgnitionContent, int, error) {
+func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string) (*isoeditor.IgnitionContent, string, int, error) {
```
carbonin, line 94: I don't like all these return values.
If someone has a better idea for how to organize this I'd be happy for a suggestion.

internal/handlers/initrd.go
```diff
@@ -68,5 +69,10 @@ func (h *initrdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	fileName := fmt.Sprintf("%s-initrd.img", imageID)
 	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", fileName))
-	http.ServeContent(w, r, fileName, time.Now(), initrdReader)
+	modTime, err := http.ParseTime(lastModified)
+	if err != nil {
+		log.Warnf("Error parsing last modified time %s: %v", lastModified, err)
+		modTime = time.Now()
```
danielerez, line 75: Maybe worth adding a negative test for this, i.e. in case of an err in ParseTime.

internal/handlers/iso.go
```diff
@@ -69,11 +69,14 @@ func (h *isoHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer isoReader.Close()
 
-	//TODO: set modified time correctly (MGMT-7274)
-
 	fileName := fmt.Sprintf("%s-discovery.iso", imageID)
 	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", fileName))
-	http.ServeContent(w, r, fileName, time.Now(), isoReader)
+	modTime, err := http.ParseTime(lastModified)
+	if err != nil {
```
danielerez, line 75: same

PR #75, opened 2022-05-02
WIP NO-ISSUE: Bump golangci-lint to 1.45.2

vrutkovs:


carbonin:
Maybe use what assisted-service is using?
https://github.com/openshift/assisted-service/blob/master/hack/setup_env.sh#L73-L74

Or is there an issue with that version too we think?

vrutkovs:
I was thinking we could update both, starting with this repo.

However, the issue is that we did `go import` which broke down - and apparently not needed for tests?

carbonin:
> However, the issue is that we did go import which broke down - and apparently not needed for tests?

Do you mean removing `go get -u golang.org/x/tools/cmd/goimports@v0.1.5 github.com/golang/mock/mockgen@v1.6.0` is what fixed it?

I think we need that for `make format` unless we're not running that in the CI?

carbonin:
If this fixes the issue and we don't need to use these commands in CI then I'm okay with removing them I guess :shrug:

vrutkovs:
> Do you mean removing go get -u golang.org/x/tools/cmd/goimports@v0.1.5 github.com/golang/mock/mockgen@v1.6.0 is what fixed it?

Right, I'll check if `make format` works without these and split this change in a different PR

vrutkovs:
Will be reworked after #76

vrutkovs:
This bump is useless since `golangci-lint` version is defined in the image :/

PR #80, opened 2022-06-01
MGMT-10623: Add OCP 4.11 nightly images

osherdp:
## Description

This change should mainly affect integration env, having nightly-built images of OCP 4.11 (no fc/rc release has been promoted to date).

I also took the liberty to add 4.10 ARM image which seems missing.

## How was this code tested?

``oc process --local -f deploy/template.yaml -pIMAGE_TAG=latest`` seems to work.
For some reason I couldn't make ``RHCOS_VERSIONS`` have a minimal value (without line-breaks), even when using ``value: <-`` but I think it's not a big issue.

## Assignees

/cc @carbonin 
/cc @danielerez 

## Links
[MGMT-10623](https://issues.redhat.com//browse/MGMT-10623)

## Checklist

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -6,7 +6,72 @@ parameters:
   value: ''
   required: true
 - name: OS_IMAGES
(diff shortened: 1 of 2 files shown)


osherdp:
> Why does this have a hold?
While checking configuration of 4.11 images in staging, I found out we can't really use those 4.11 nightly OCP releases with staging pull-secret. It means we can push those ISO images, but either way cannot use any of it for a full-installation flow. I'm considering waiting for the fc-0 release instead of having those problematic releases
Prod pull-secret works just fine with those nightly releases btw

osherdp:
/unhold
4.11.0-fc.0 is available now, so no real reason to hold this one
PTAL @carbonin @danielerez

deploy/template.yaml
```diff
@@ -6,7 +6,72 @@ parameters:
...
+        "openshift_version": "4.10",
+        "cpu_architecture": "arm64",
+        "url": "https://mirror.openshift.com/pub/openshift-v4/aarch64/dependencies/rhcos/4.10/4.10.3/rhcos-4.10.3-aarch64-live.aarch64.iso",
+        "rootfs_url": "https://mirror.openshift.com/pub/openshift-v4/aarch64/dependencies/rhcos/4.10/4.10.3/rhcos-4.10.3-aarch64-live-rootfs.aarch64.img",
+        "version": "410.84.202201251210-0"
+      },
+      {
+        "openshift_version": "4.11",
+        "cpu_architecture": "x86_64",
+        "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.11/rhcos-live.x86_64.iso",
```
carbonin, line 63: This doesn't point to a specific image.
This means that it might be difficult to force an upgrade of this image.

Would it be better to pin to a particular image link like this one?
https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/4.11.0-0.nightly-2022-04-16-163450/

osherdp: We had this discussion here: https://github.com/openshift/assisted-service/pull/3862#discussion_r884284295
I checked the history from back when 4.10 got in, and it seems like pinned nightly images were removed after a while (see https://github.com/openshift/assisted-service/commit/94ce200b59e8bb3fa0032819a727aa85b532dc22)

If we can pin it to a particular image that is also persistent then it will be better

carbonin: That's what I'm suggesting.
Look at my comment from the problem PR https://github.com/openshift/assisted-service/pull/3196#discussion_r785057710 and @danielerez comment in the new one https://github.com/openshift/assisted-service/pull/3862#discussion_r884682147

The problematic link is `/pre-release/latest-4.11/<some-specific-image>`
We're suggesting `/pre-release/<specific-version>/<specific-image>`

osherdp: Done
Please take another look

carbonin (approved):

PR #83, opened 2022-06-15
MGMT-10751: use boot-artifacts API as rootfs URL

danielerez:
## Description

When creating a minimal ISO template the rootfsURL is injected to grub config files.
For each OS image metadata (OS_IMAGES) the rootfs_url is available as an external URL to the img file.
This URL is currently used on imagestore -> Populate.

This change is removing the requirement to specify an external rootfs URL for each image, instead, we're now using the
/boot-artifacts/rootfs API. I.e. a URL as follows is generated when creating the minimal ISO and injected to the grub configs:
http://`<image-service-base-url>`/boot-artifacts/rootfs?arch=x86_64&version=4.10

Hence, when deploying the image service, IMAGE_SERVICE_BASE_URL env var should be now specified.

Note: the previous behaviour of using the rootfs_url from OS_IMAGES is kept as a fallback for now. Can be removed later on once ensuring IMAGE_SERVICE_BASE_URL is available for the deployment.

## How was this code tested?
Tested minimal ISO functionality manually and modified/added relevant unit tests.

## Assignees
/cc @carbonin 

## Checklist

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 
@@ -69,10 +70,12 @@ type ImageStore interface {
 }
 
 type rhcosStore struct {
-	versions   []map[string]string
-	isoEditor  isoeditor.Editor
-	dataDir    string
-	httpClient *http.Client
+	versions           []map[string]string
+	isoEditor          isoeditor.Editor
+	dataDir            string
+	httpClient         *http.Client
+	imageServiceScheme string
+	imageServiceHost   string
 }
 
 const (
@@ -80,7 +83,7 @@ const (
 	ImageTypeMinimal = "minimal-iso"
 )
 
-func NewImageStore(ed isoeditor.Editor, dataDir string, insecureSkipVerify bool, versions []map[string]string) (ImageStore, error) {
+func NewImageStore(ed isoeditor.Editor, dataDir, imageServiceScheme, imageServiceHost string, insecureSkipVerify bool, versions []map[string]string) (ImageStore, error) {
 	if err := validateVersions(versions); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -209,7 +213,17 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 			log.Infof("Creating minimal iso for %s-%s-%s", openshiftVersion, imageVersion, arch)
 
 			fullPath := filepath.Join(s.dataDir, isoFileName(ImageTypeFull, openshiftVersion, imageVersion, arch))
-			err = s.isoEditor.CreateMinimalISOTemplate(fullPath, imageInfo["rootfs_url"], minimalPath)
+			rootfsURL, err := buildURL(s.imageServiceBaseURL, fmt.Sprintf("%s/rootfs", "/boot-artifacts"), map[string]string{
```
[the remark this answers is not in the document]
carbonin: Also I feel like the priority of this would be more clear if we did something like this rather than erroring out `buildURL` if we pass an empty base URL.

```go
rootfsURL := imageInfo["rootfs_url"]
if s.imageServiceBaseURL != "" {
  rootfsURL, err = buildURL(...)
  if err != nil {
    log.Errorf("...")
    return err
  }
}
```

This way we will only fallback if the base url is empty vs when we set it but it failed to parse or something.

carbonin (commented):
I guess we'll also remove the validation that rootfs is present in the version in a followup?

pkg/imagestore/imagestore.go
```diff
@@ -235,6 +252,26 @@ func isoFileName(imageType, openshiftVersion, version, arch string) string {
 	return fmt.Sprintf("rhcos-%s-%s-%s-%s.iso", imageType, openshiftVersion, version, arch)
 }
 
+func buildRootfsURL(baseURL string, suffix string, params map[string]string) (string, error) {
+	base, err := url.Parse(baseURL)
+	if err != nil || baseURL == "" {
```
carbonin, line 257: You shouldn't have the check `baseURL` for empty here now, right?

pkg/imagestore/imagestore.go
```diff
@@ -235,6 +244,10 @@ func isoFileName(imageType, openshiftVersion, version, arch string) string {
 	return fmt.Sprintf("rhcos-%s-%s-%s-%s.iso", imageType, openshiftVersion, version, arch)
 }
 
+func buildRootfsURL(baseURL string, arch, version string) string {
+	return fmt.Sprintf("%s/boot-artifacts/rootfs?arch=%s&version=%s", baseURL, arch, version)
```
carbonin, line 248: Why remove the URL builder? That will handle things like if someone puts a slash at the end of the url or if the version contains url unsafe characters for some reason.

danielerez: yeah good point, added.

carbonin (approved):

PR #85, opened 2022-07-05
MGMT-11048: add image_service_host var to template

danielerez:
## Description
* As the service CRD for SaaS deployment is in the template, added IMAGE_SERVICE_HOST env var.
* As we can't fetch the external url of the image-service, Set IMAGE_SERVICE_HOST by IMAGE_SERVICE_BASE_URL env var.
* In buildRootfsURL, added support for a host string that contains scheme and path.

## How was this code tested?
Added relevant unit-tests.

## Assignees
/cc @carbonin 

## Checklist

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -77,6 +77,9 @@ parameters:
   value: 15Gi
 - name: REPLICAS_COUNT
   value: "3"
+- name: IMAGE_SERVICE_BASE_URL
+  value: ''
+  required: true
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
@@ -123,6 +126,8 @@ objects:
             value: http
           - name: ASSISTED_SERVICE_HOST
             value: "assisted-service:8090"
+          - name: IMAGE_SERVICE_HOST
+            value: ${IMAGE_SERVICE_BASE_URL}
           - name: DATA_DIR
             value: "/data"
           ports:
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -8,7 +8,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/renameio"
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
@@ -233,13 +235,16 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 
 			fullPath := filepath.Join(s.dataDir, isoFileName(ImageTypeFull, openshiftVersion, imageVersion, arch))
 			var rootfsURL string
-			if s.imageServiceScheme == "" || s.imageServiceHost == "" {
+			if s.imageServiceHost == "" {
 				if _, ok := imageInfo["rootfs_url"]; !ok {
 					return fmt.Errorf("invalid version entry %+v: missing rootfs_url key", imageInfo)
 				}
 				rootfsURL = imageInfo["rootfs_url"]
 			} else {
-				rootfsURL = buildRootfsURL(s.imageServiceScheme, s.imageServiceHost, arch, openshiftVersion)
+				rootfsURL, err = buildRootfsURL(s.imageServiceScheme, s.imageServiceHost, arch, openshiftVersion)
+				if err != nil {
+					return fmt.Errorf("failed to build rootfs URL: %v", err)
+				}
 			}
 
 			err = s.isoEditor.CreateMinimalISOTemplate(fullPath, rootfsURL, minimalPath)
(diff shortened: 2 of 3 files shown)


deploy/template.yaml
```diff
@@ -123,6 +123,10 @@ objects:
             value: http
           - name: ASSISTED_SERVICE_HOST
             value: "assisted-service:8090"
+          - name: IMAGE_SERVICE_SCHEME
+            value: https
+          - name: IMAGE_SERVICE_HOST
+            value: api.integration.openshift.com/api/assisted-images
```
carbonin, line 129: Now that we're parsing this (to get the path) does it make sense to allow someone to specify the scheme as a part of this env var?

That would allow us to specify this as `IMAGE_SERVICE_HOST: ${OCM_BASE_URL}/api/assisted-images` in the saas file and not also specify the scheme.

We would need to implement some priority logic in the source to deal with if someone specifies `IMAGE_SERVICE_SCHEME` and also a scheme in the host env var, but that shouldn't be too bad.

Of course this will devolve into us not really needing the scheme env var, but I don't know that we want to go back and change all the other deployment methods now :tired_face: 

WDYT @osherdp @gamli75 @danielerez ?

danielerez: It won't be too clean to allow scheme in the host, but would indeed simplify the changes in saas file.
So maybe worth the change. I guess we could clean up the scheme var later from other deployments. I'll update the PR just to support it and we'll see how it goes.

carbonin: That all said, should we use the full URL here instead of splitting?

Also, this needs to be configurable. So list this in the parameters, put the default there and use the parameter here.

carbonin (changes requested):

deploy/template.yaml
osherdp: What's written as IMAGE_SERVICE_HOST is not really the host part but rather host+path (sometimes called also URI)
I think we should have the full URL in one var, it will be easier to manage and work with

carbonin: Yeah, that's the issue we're dealing with here.
We forgot that `OCM_BASE_URL` and the additional path was something that we were going to have to deal with when deploying to the SaaS. So we're trying to deal with that while also not changing how we're deploying everywhere else.

osherdp (changes requested):

deploy/template.yaml
```diff
@@ -77,6 +77,8 @@ parameters:
   value: 15Gi
 - name: REPLICAS_COUNT
   value: "3"
+- name: IMAGE_SERVICE_HOST
```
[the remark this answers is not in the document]
danielerez: Yeah I guess by this point we can call it IMAGE_SERVICE_BASE_URL (as in assisted-service)

deploy/template.yaml
```diff
@@ -77,6 +77,8 @@ parameters:
   value: 15Gi
 - name: REPLICAS_COUNT
   value: "3"
+- name: IMAGE_SERVICE_BASE_URL
+  value: https://api.integration.openshift.com/api/assisted-images
```
carbonin, line 81: I think this should probably be empty and required like we do in [assisted service](https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml#L8-L10)

That would mean adding the parameter to the saas file first though.

pkg/imagestore/imagestore.go
```diff
@@ -268,17 +272,24 @@ func isoFileName(imageType, openshiftVersion, version, arch string) string {
...
 }
 
-func buildRootfsURL(scheme, host, arch, version string) string {
+func buildRootfsURL(scheme, host, arch, version string) (string, error) {
+	base, err := url.Parse(host)
+	if err != nil {
+		return "", err
+	}
+	if scheme == "" {
+		scheme = base.Scheme
```
carbonin, line 281: Should we have another fallback here if the base url doesn't have a scheme? Would `url.Parse` error in that case?

carbonin: In a quick test I did it looks like if there is no scheme the Parse won't fail, but the host gets put into the path so we probably want to watch for that.

danielerez: So worth checking it explicitly or add a comment about it?

carbonin: I think explicitly.

danielerez: Updated. Looks a bit cumbersome though:/

osherdp (changes requested):

deploy/template.yaml
```diff
@@ -123,6 +126,8 @@ objects:
             value: http
           - name: ASSISTED_SERVICE_HOST
             value: "assisted-service:8090"
+          - name: IMAGE_SERVICE_HOST
```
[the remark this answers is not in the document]
danielerez: that's the var we use in the code: https://github.com/openshift/assisted-image-service/blob/f6af8ed8e6e6c1f6df0ba4a0301c488d620ab007/main.go#L39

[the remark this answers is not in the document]
osherdp: yeah I was wondering why we need to have a difference between those two names (because I do like better the external name)
but if that might be a problem for backwards-compatibility then no need to change it

osherdp:
Basically looks good to me, but the ``deploy-nodes`` job is failing. Is it connected to those changes?

osherdp (approved):

danielerez:
> Basically looks good to me, but the `deploy-nodes` job is failing. Is it connected to those changes?

Right seems related. Fixed, thanks!

carbonin:
I'm not going to hold this up any more, but I do really think we need some follow-up work to standardize how we're providing this url and what kind of values we accept.

Ideally at this point we should expect a url, and only a url that parses correctly, and remove the scheme env var.

pkg/imagestore/imagestore.go
```diff
@@ -268,17 +273,45 @@ func isoFileName(imageType, openshiftVersion, version, arch string) string {
 	return fmt.Sprintf("rhcos-%s-%s-%s-%s.iso", imageType, openshiftVersion, version, arch)
 }
 
-func buildRootfsURL(scheme, host, arch, version string) string {
+func buildRootfsURL(scheme, host, arch, version string) (string, error) {
+	if strings.Contains(host, ":") && !strings.Contains(host, "http") {
+		// url can't parse host with colon when scheme is missing
+		host = fmt.Sprintf("%s://%s", scheme, host)
```
carbonin, line 279: This assumes scheme is passed.

pkg/imagestore/imagestore.go
```diff
@@ -268,17 +273,45 @@ func isoFileName(imageType, openshiftVersion, version, arch string) string {
 	return fmt.Sprintf("rhcos-%s-%s-%s-%s.iso", imageType, openshiftVersion, version, arch)
 }
 
-func buildRootfsURL(scheme, host, arch, version string) string {
+func buildRootfsURL(scheme, host, arch, version string) (string, error) {
+	if strings.Contains(host, ":") && !strings.Contains(host, "http") {
```
carbonin, line 277: Should probably check for host starting with http rather than contains.

danielerez: Right, changed to HasPrefix

pkg/imagestore/imagestore.go
danielerez: Yeah, so in this case we'll get 'missing protocol scheme' error, which should be fine for now I guess.

carbonin (approved):

carbonin:
/unhold

I'm okay with this.

It's a step in the right direction (supporting full urls) even if we need some more work to standardize on that approach.

danielerez:
> /unhold
> 
> I'm okay with this.
> 
> It's a step in the right direction (supporting full urls) even if we need some more work to standardize on that approach.

Opened this a a follow-up: https://issues.redhat.com/browse/MGMT-11124

PR #88, opened 2022-07-18
MGMT-10957: remove fallback for missing imageServiceHost

danielerez:
## Description
* As all deployments should already specify IMAGE_SERVICE_SCHEME/HOST, removed fallback behaviour when the env vars are missing.
* Removed 'rootfs_url' from all OS image entities.
* Added to Makefile default values of IMAGE_SERVICE_SCHEME/HOST for running the service locally.
* To follow: support IMAGE_SERVICE_BASE_URL env var as a preparation for removing IMAGE_SERVICE_SCHEME.

## How was this code tested?
Updated tests and verified on local env.

## Assignees
/cc @carbonin 

## Checklist


danielerez:
> /hold
> 
> This looks good, but I think we're still frozen, right?

Yeah thanks, will unhold when possible.

PR #93, opened 2022-08-25
MGMT-11442: Fetch discovery ignition with static networking if minimal iso requested.

paul-maidment:
This will make a request to the assisted-service, passing the value of the request parameter "imageType"
When the PR https://github.com/openshift/assisted-service/pull/4309 is merged then the assisted service will respond by
including static networking config in the discovery ignition when a minimal iso is requested.

The static networking config will not be included in the ignition if a full iso is requested.

## How was this code tested?
This code is covered by modification of unit tests to cover the functionality.

## Assignees

/cc @carbonin 
/cc @nmagnezi 

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -111,9 +111,10 @@ var _ = Describe("Image integration tests", func() {
 				assistedServer = ghttp.NewServer()
 				u, err := url.Parse(assistedServer.URL())
 				Expect(err).NotTo(HaveOccurred())
+				queryString := fmt.Sprintf("file_name=discovery.ign&discovery_iso_type=%s", tc.imageType)
 				assistedServer.AppendHandlers(
 					ghttp.CombineHandlers(
-						ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/files", imageID), "file_name=discovery.ign"),
+						ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/files", imageID), queryString),
 						ghttp.RespondWith(http.StatusOK, tc.expectedIgnition),
 					),
 				)
--- a/internal/handlers/iso.go
+++ b/internal/handlers/iso.go
@@ -44,7 +44,7 @@ func (h *isoHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ignition, lastModified, statusCode, err := h.client.ignitionContent(r, imageID)
+	ignition, lastModified, statusCode, err := h.client.ignitionContent(r, imageID, params.imageType)
 	if err != nil {
 		log.Errorf("Error retrieving ignition content: %v\n", err)
 		w.WriteHeader(statusCode)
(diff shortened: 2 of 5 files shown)


internal/handlers/iso_test.go
```diff
@@ -136,6 +136,12 @@ var _ = Describe("ServeHTTP", func() {
 			})
 
 			It("returns a full image", func() {
+				assistedServer.SetHandler(0,
```
carbonin, line 139: Maybe remove the `AppendHandlers` call from `BeforeEach` and create a helper for expecting a call to the assisted server.
That should allow you to deduplicate all of this while also expecting the different iso types.

internal/handlers/iso_test.go
```diff
@@ -91,6 +91,15 @@ var _ = Describe("ServeHTTP", func() {
 			Expect(respContent).To(Equal(content))
 		}
 
+		initISOHandler := func(queryString string) {
```
carbonin, line 94: Nit, but this is a handler for the ignition, not the iso.

internal/handlers/iso_test.go
```diff
@@ -189,13 +196,15 @@ var _ = Describe("ServeHTTP", func() {
 			})
 
 			It("fails when no version is supplied", func() {
+				initISOHandler("file_name=discovery.ign")
```
carbonin, line 199: I don't think you actually need this for the failure cases.

carbonin (commented):
Looks good, just a few minor test comments.

internal/handlers/iso.go
```diff
@@ -44,7 +44,7 @@ func (h *isoHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ignition, lastModified, statusCode, err := h.client.ignitionContent(r, imageID)
+	ignition, lastModified, statusCode, err := h.client.ignitionContent(r, imageID, params.imageType)
```
nmagnezi, line 47: Won't the older versions of `assisted-service` reject this request?
If I'm right, we should probably catch this response and make sure we log something meaningful for users about these circumstances.

paul-maidment: Something along the lines of "The version of assisted service in use does not permit override of discovery_iso_type, if you wish to generate the ISO based on a specific type, you will need to edit the infraenv" ?

paul-maidment: It's perfectly valid to make a request such as 

`curl   --silent   --header "Authorization: Bearer $(ocm token)"   "https://api.openshift.com/api/assisted-install/v2/infra-envs/9f49dd92-80b7-4995-84f5-2fac5b535cb3/downloads/files?file_name=discovery.ign&somerandomparam=1" | jq`

In the example above, somerandomparam will be ignored unless it is supported by assisted-service.

This demonstrates that the change is backwards compatible I think?

integration_test/images_test.go
```diff
@@ -76,24 +76,28 @@ var _ = Describe("Image integration tests", func() {
 
 	testcases := []struct {
 		name             string
+		expectedIsoType  string
```
carbonin, line 79: Why is this different from `imageType`?

PR #94, opened 2022-08-25
NO-ISSUE: Performance enhancements for unit tests

paul-maidment:
## Description
When running unit tests locally, I ran into a couple of issues

1: The tests would have frequent connection failures when trying to create http servers on various ports, this felt like a bug, but would not consistently happen for all users.
2: The unit tests would take a long time to "clean up" adding a lot of uneccesary overhead to tests.

This commit makes sure that when we launch a HTTP or HTTPS server that we wait a reasonable amount of time for them to come online before exiting the non-blocking ListenAndServe() method.
We also added a "Fast shutdown" mode to the HTTP and HTTPS servers where the connection will be closed instead of shut down. This is exclusively for use in testing environments and should not be used in production.

## How was this code tested?
By running the unit tests

## Assignees

/cc @carbonin 
/cc @nmagnezi 

## Checklist

### DIFF
--- a/pkg/servers/servers.go
+++ b/pkg/servers/servers.go
@@ -14,6 +14,7 @@ type ServerInfo struct {
 	HTTPSKeyFile    string
 	HTTPSCertFile   string
 	HasBothHandlers bool
+	FastShutdown    bool
 }
 
 func New(httpPort, httpsPort, HTTPSKeyFile, HTTPSCertFile string) *ServerInfo {
@@ -58,6 +59,7 @@ func (s *ServerInfo) ListenAndServe() {
 	if s.HTTP != nil {
 		go s.httpListen()
 	}
+
 	if s.HTTPS != nil {
 		go s.httpsListen()
 	}
@@ -65,10 +67,18 @@ func (s *ServerInfo) ListenAndServe() {
 
 func (s *ServerInfo) Shutdown() bool {
 	if s.HTTPS != nil {
-		shutdown("HTTPS", s.HTTPS)
+		if s.FastShutdown {
+			s.HTTPS.Close()
+		} else {
+			shutdown("HTTPS", s.HTTPS)
+		}
 	}
 	if s.HTTP != nil {
-		shutdown("HTTP", s.HTTP)
+		if s.FastShutdown {
+			s.HTTP.Close()
+		} else {
+			shutdown("HTTP", s.HTTP)
+		}
 	}
 	return true
 }
--- a/pkg/servers/servers_test.go
+++ b/pkg/servers/servers_test.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
@@ -26,6 +27,16 @@ var httpsKeyFile, httpsCertFile *os.File
 var ready func(rw http.ResponseWriter, req *http.Request)
 var mux *http.ServeMux
 var httpClient, httpsClient *http.Client
+
+const portConnectionRetrySeconds = 30
+const portConnectionRetryInterval = 10 * time.Millisecond
+
+// Create a new instance of the server under test
+var NewServer = func(httpPort, httpsPort, HTTPSKeyFile, HTTPSCertFile string) *ServerInfo {
+	server := New(httpPort, httpsPort, HTTPSKeyFile, HTTPSCertFile)
+	server.FastShutdown = true
+	return server
+}
 
 var _ = BeforeSuite(func() {
 	var err error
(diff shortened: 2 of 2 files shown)


pkg/servers/servers.go
```diff
@@ -54,21 +56,51 @@ func shutdown(name string, server *http.Server) {
...
+		}
+	}
+
+	return nil
 }
 
 func (s *ServerInfo) Shutdown() bool {
 	if s.HTTPS != nil {
-		shutdown("HTTPS", s.HTTPS)
+		go shutdown("HTTPS", s.HTTPS)
```
carbonin, line 100: No so much a fan of this. I'd prefer to somehow tell this function to shutdown more quickly just for unit tests.

But I also don't want to deal with changing the signature as that would mean a change to assisted-service as well

carbonin:
It would also be good to add a description to the PR about why you're making this change.

filanov (commented):
few comments, maybe some of them are not related directly to this PR. 

1. Why use static ports in tests? why not use https://pkg.go.dev/net/http/httptest ?
2. if retry is used only in tests why not to use eventaully in the test case? or use awaitConnection as a helper function for the tests in each `BeforeEach`

paul-maidment:
> 1. Why use static ports in tests? why not use https://pkg.go.dev/net/http/httptest ?

Mainly because the HTTP or HTTPS server is created by the code under test in this case. The HTTP server is not a "test" one but is meant to represent the "production" server created in code. Though I do see some value in acquiring a "free port" and using this in the test.

paul-maidment:
> few comments, maybe some of them are not related directly to this PR.
> 
>     1. Why use static ports in tests? why not use https://pkg.go.dev/net/http/httptest ?
> 
>     2. if retry is used only in tests why not to use eventaully in the test case? or use awaitConnection as a helper function for the tests in each `BeforeEach`

Agreed, I will change this, keeps test code in the tests.

filanov:
I agree that finding a free port is better then having it hard coded, but there is a risk here as well because unit tests are usually running in parallel

pkg/servers/servers_test.go
```diff
@@ -27,6 +29,17 @@ var ready func(rw http.ResponseWriter, req *http.Request)
 var mux *http.ServeMux
 var httpClient, httpsClient *http.Client
 
+func awaitConnection(address string, numberOfRetries int, retryInterval time.Duration) error {
```
filanov, line 32: can be replaced with eventually
for example https://github.com/openshift/assisted-service/blob/master/subsystem/kubeapi_test.go#L117

carbonin:
cc @vrutkovs

pkg/servers/servers.go
```diff
@@ -14,9 +14,10 @@ type ServerInfo struct {
 	HTTPSKeyFile    string
 	HTTPSCertFile   string
 	HasBothHandlers bool
+	FastShutdown    bool
 }
 
-func New(httpPort, httpsPort, HTTPSKeyFile, HTTPSCertFile string) *ServerInfo {
+func New(httpPort, httpsPort, HTTPSKeyFile, HTTPSCertFile string, fastShutdown bool) *ServerInfo {
```
carbonin, line 20: Changing this will require a change in the assisted service when we pull in the new code.

If we're only adding `fastShutdown` for tests I think it's probably better to not change this function signature and set the `FastShutdown` struct field directly in the tests.

carbonin (changes requested):

pkg/servers/servers.go
vrutkovs: +1 to that, we don't want this to be configurable in production deployment

pkg/servers/servers_test.go
```diff
@@ -89,54 +89,55 @@ var _ = AfterSuite(func() {
...
 		Expect(listeners.HTTP.Addr).To(Equal(":8080"))
 		Expect(listeners.HTTPS).To(BeNil())
 		Expect(listeners.HasBothHandlers).To(BeFalse())
 
 		listeners.ListenAndServe()
-		timeout := time.Second
-		_, err := net.DialTimeout("tcp", net.JoinHostPort("localhost", "8080"), timeout)
-		Expect(err).NotTo(HaveOccurred())
+		Eventually(func() error {
+			_, err := net.DialTimeout("tcp", "localhost:8080", 100*time.Millisecond)
```
vrutkovs, line 101: Lets add `100*time.Millisecond` as a const. Improving the connection setup in unittests up to 10 is good, lets see how far can we push it?

vrutkovs (approved):
/lgtm

Nick, could you have another look and cancel the hold?

pkg/servers/servers_test.go
```diff
@@ -27,6 +28,16 @@ var ready func(rw http.ResponseWriter, req *http.Request)
 var mux *http.ServeMux
 var httpClient, httpsClient *http.Client
 
+const RETRY_SECONDS = 30
+const RETRY_INTERVAL = 10 * time.Millisecond
```
carbonin, line 32: Nit, but I don't thing we generally define consts in caps and I think they should be named specifically to where they're used (for connection retry).

carbonin (approved):

PR #97, opened 2022-09-30
MGMT-12189: Include static network config in initrd pxe artifact

carbonin:
## Description

Originally it was thought that this would not be necessary as iPXE requires network access to work. This became an issue when working with ironic though.

In order to configure iPXE for the hosts it is trying to boot ironic controlls DHCP, and therefore DNS, for those hosts. The problem is that they don't forward DNS requests to any DNS servers that happen to already be configured in the environment.

Adding the static networking config to the initrd artifact will enable a workaround for this issue where the user would provide DNS configuration to the hosts using nmstate even if that configuration was already present in the environment.

## How was this code tested?

Built and applied this patch to an environment with the ironic issue described in :point_up: 
Applied an nmstate configuration to the infraEnv and the host was able to successfully pull the rootfs.

## Assignees

/cc @CrystalChun 
/cc @danielerez 

## Links

https://issues.redhat.com/browse/MGMT-12189
related ironic issue: https://issues.redhat.com/browse/OCPBUGS-1844

## Checklist

### DIFF
--- a/internal/handlers/initrd.go
+++ b/internal/handlers/initrd.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/handlers/initrd.go
```diff
@@ -67,6 +68,21 @@ func (h *initrdHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer initrdReader.Close()
 
+	ramdisk, statusCode, err := h.client.ramdiskContent(r, imageID)
```
CrystalChun, line 71: Sorry, I might be missing something, but won't this get all of the minimal initrd as opposed to just the static network config files?

carbonin: This could all use better names at this point.

This will get the initrd from assisted which contains the static network config files. The same one that we embed in the minimal iso.

You can see how it's created [here](https://github.com/openshift/assisted-service/blob/e69c126e90b94eff81a193a2b8fd132451971bd2/internal/bminventory/inventory.go#L3381-L3415). It's pretty much just the static network config.

CrystalChun: Gotcha, thanks for the clarification!

PR #99, opened 2022-10-23
MGMT-12054: Support discovery kernel arguments for full-iso and minimal-iso

ori-amizur:
When streaming the discovery ISO file, the kernel arguments are included as part of some files: grub.cfg and isolinux.cfg

Each of these files contains an area where the kernel arguments can be embedded. This are can be identified by the regex "(\n#+) COREOS_KARG_EMBED_AREA". The arguments should overwrite this are from the beginning and should not pass the last # (hash sign). To do that, an overlay was created.

In new Openshift versions, there is a file that contains the file names, while in old versions the file does not exist. If the file exists, the files to embed the kernel arguments are taken from this config file. Otherwise, the default files [/EFI/redhat/grub.cfg, /isolinux/isolinux.cfg] are used.

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

### DIFF
--- a/internal/handlers/iso_test.go
+++ b/internal/handlers/iso_test.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -8,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo"
@@ -100,6 +102,39 @@ var _ = Describe("ServeHTTP", func() {
 			)
 		}
 
+		setInfraenvKargsHandlerSuccess := func(args ...string) {
+			var response string
+			if len(args) == 0 {
+				response = "{}"
+			} else {
+				discoveryKernelArguments, err := isoeditor.KargsToStr(args)
+				Expect(err).ToNot(HaveOccurred())
+				var infraEnv struct {
+					// JSON formatted string array representing the discovery image kernel arguments.
+					DiscoveryKernelArguments *string `json:"discovery_kernel_arguments,omitempty"`
+				}
+				infraEnv.DiscoveryKernelArguments = &discoveryKernelArguments
+				b, err := json.Marshal(&infraEnv)
+				Expect(err).ToNot(HaveOccurred())
+				response = string(b)
+			}
+			assistedServer.AppendHandlers(
+				ghttp.CombineHandlers(
+					ghttp.VerifyRequest("GET", fmt.Sprintf(infraEnvPathFormat, imageID)),
+					ghttp.RespondWith(http.StatusOK, response, header),
+				),
+			)
+		}
+
+		setInfraenvKargsHandlerFailure := func() {
+			assistedServer.AppendHandlers(
+				ghttp.CombineHandlers(
+					ghttp.VerifyRequest("GET", fmt.Sprintf(infraEnvPathFormat, imageID)),
+					ghttp.RespondWith(http.StatusInternalServerError, "", header),
+				),
+			)
+		}
+
 		Context("with no auth", func() {
 			var (
 				server        *httptest.Server
@@ -112,7 +147,7 @@ var _ = Describe("ServeHTTP", func() {
 				Expect(err).NotTo(HaveOccurred())
 
 				initrdContent = nil
-				mockImageStream := func(isoPath string, ignition *isoeditor.IgnitionContent, ramdiskBytes []byte) (isoeditor.ImageReader, error) {
+				mockImageStream := func(isoPath string, ignition *isoeditor.IgnitionContent, ramdiskBytes, kargs []byte) (isoeditor.ImageReader, error) {
 					defer GinkgoRecover()
 					Expect(ignition.Config).To(Equal([]byte(ignitionContent)))
 					if isoPath == minImageFilename {
@@ -141,6 +176,7 @@ var _ = Describe("ServeHTTP", func() {
 				initIgnitionHandler("discovery_iso_type=full-iso&file_name=discovery.ign")
 				mockImage("4.8", imagestore.ImageTypeFull, defaultArch)
 				path := fmt.Sprintf("/images/%s?version=4.8&type=full-iso", imageID)
+				setInfraenvKargsHandlerSuccess()
 				resp, err := client.Get(server.URL + path)
 				Expect(err).NotTo(HaveOccurred())
 				expectSuccessfulResponse(resp, []byte("someisocontent"))
@@ -150,6 +186,7 @@ var _ = Describe("ServeHTTP", func() {
 				initIgnitionHandler("discovery_iso_type=full-iso&file_name=discovery.ign")
 				mockImage("4.8", imagestore.ImageTypeFull, "arm64")
 				path := fmt.Sprintf("/images/%s?version=4.8&type=full-iso&arch=arm64", imageID)
+				setInfraenvKargsHandlerSuccess()
 				resp, err := client.Get(server.URL + path)
 				Expect(err).NotTo(HaveOccurred())
 				expectSuccessfulResponse(resp, []byte("someisocontent"))
@@ -164,6 +201,7 @@ var _ = Describe("ServeHTTP", func() {
 						ghttp.RespondWith(http.StatusOK, initrdContent),
 					),
 				)
+				setInfraenvKargsHandlerSuccess()
 				mockImage("4.8", imagestore.ImageTypeMinimal, defaultArch)
 				path := fmt.Sprintf("/images/%s?version=4.8&type=minimal-iso", imageID)
 				resp, err := client.Get(server.URL + path)
@@ -181,6 +219,7 @@ var _ = Describe("ServeHTTP", func() {
 				)
 				mockImage("4.8", imagestore.ImageTypeMinimal, defaultArch)
 				path := fmt.Sprintf("/images/%s?version=4.8&type=minimal-iso", imageID)
+				setInfraenvKargsHandlerSuccess()
 				resp, err := client.Get(server.URL + path)
 				Expect(err).NotTo(HaveOccurred())
 				expectSuccessfulResponse(resp, []byte("minimalisocontent"))
@@ -221,9 +260,20 @@ var _ = Describe("ServeHTTP", func() {
 				header.Set("Last-Modified", "somenonsense")
 				mockImage("4.8", imagestore.ImageTypeFull, defaultArch)
 				path := fmt.Sprintf("/images/%s?version=4.8&type=full-iso", imageID)
+				setInfraenvKargsHandlerSuccess()
 				resp, err := client.Get(server.URL + path)
 				Expect(err).NotTo(HaveOccurred())
 				expectSuccessfulResponse(resp, []byte("someisocontent"))
+			})
+
+			It("fail kargs infra env query", func() {
+				initIgnitionHandler("discovery_iso_type=full-iso&file_name=discovery.ign")
+				mockImage("4.8", imagestore.ImageTypeFull, defaultArch)
+				path := fmt.Sprintf("/images/%s?version=4.8&type=full-iso", imageID)
+				setInfraenvKargsHandlerFailure()
+				resp, err := client.Get(server.URL + path)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(resp.StatusCode).To(Equal(http.StatusInternalServerError))
 			})
 		})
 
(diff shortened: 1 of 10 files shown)


cgwalters:
Ahhh wait, is there a reason this code isn't using `coreos-installer`?  I am not totally sure our existing kargs area is stable public API.

carbonin:
@cgwalters I think there are still a few gaps.

The last feature gap IIRC was that the minimal ISO generated by coreos-installer doesn't let us set up our static networking (posisibly for multiple interfaces) and proxy information before pulling the rootfs.

I recall talking to @jlebon about this a while back

carbonin:
But it may very well be time to meet up again to discuss if everything is covered now.
I definitely wouldn't mind moving to coreos-installer if it covered all our use cases.

carbonin:
hmmm also even if `coreos-installer` supports streaming the output iso to stdout we'd still be missing range request support I think.

I think it would be good to move this discussion out of this PR though. I'll open an issue for us to consider `coreos-installer`.

jlebon:
Yeah, we discussed this in the past. The use of `COREOS_KARG_EMBED_AREA` here is not ideal, and we're hoping that coreos-installer can satisfy AI use cases in the future. We have support for network configuration now (see https://coreos.github.io/coreos-installer/cmd/iso/#coreos-installer-iso-network-embed), but the proxy bit is still missing (tracked at https://github.com/coreos/coreos-installer/issues/794). Would be good to discuss the range request bit in more details somewhere.

carbonin:
Opened https://issues.redhat.com/browse/MGMT-12340 to track this.
I tried to add a much as I could think of to that. Let's move the discussion there

pkg/isoeditor/stream.go
```diff
@@ -39,11 +41,24 @@ func NewRHCOSStreamReader(isoPath string, ignitionContent *IgnitionContent, ramd
...
+			}
+		}
+	}
+
 	return r, nil
 }
 
-func readerForContent(isoPath string, filePath string, base io.ReadSeeker, contentReader *bytes.Reader) (overlay.OverlayReader, error) {
-	start, length, err := GetISOFileInfo(filePath, isoPath)
+func readerForContentWithBoundariesFinder(isoPath, filePath string, base io.ReadSeeker, contentReader *bytes.Reader, boundariesFinder BoundariesFinder) (overlay.OverlayReader, error) {
```
carbonin, line 60: I think just change `readerForContent` to be the generic version.
If you want to eliminate the `boundariesFinder` function here you could also find the start and length in the caller. That feels cleaner than passing in a function type to a helper.

pkg/isoeditor/stream_test.go
```diff
@@ -68,7 +68,7 @@ var _ = Describe("NewRHCOSStreamReader", func() {
```
carbonin, line 68: Any reason not to add a test case for the karg case here?

carbonin (changes requested):

pkg/isoeditor/isoutil.go
```diff
@@ -226,7 +227,7 @@ func VolumeIdentifier(isoPath string) (string, error) {
 	return strings.TrimSpace(string(volumeId)), nil
 }
 
-func GetISOFileInfo(filePath, isoPath string) (int64, int64, error) {
+func GetISOFileInfo(isoPath, filePath string) (int64, int64, error) {
```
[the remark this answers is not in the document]
ori-amizur: This order is different in all other calls even in this file.
I think it causes confusion.

pkg/isoeditor/isoutil.go
```diff
@@ -265,3 +267,17 @@ func GetFileFromISO(isoPath, filePath string) (filesystem.File, error) {
...
 }
+
+// Reads a whole specific file from the ISO image
+func ReadFileFromISO(isoPath, filePath string) ([]byte, error) {
+	f, err := GetFileFromISO(isoPath, filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	ret, err := ioutil.ReadAll(f)
```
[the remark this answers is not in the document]
ori-amizur: OK

pkg/isoeditor/kargs.go
```diff
@@ -0,0 +1,76 @@
...
+		// default is returned
+		return []string{DefaultGrubFilePath, DefaultIsolinuxFilePath}, nil
+	}
+	var kargsConfig struct {
+		Files []struct {
+			Path *string
+		}
+	}
+	if err := json.Unmarshal(kargsData, &kargsConfig); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
ori-amizur: In 4.8 the config file doesn't exist.  Therefore, fallback to the default files is correct.
When the config file exists, failing to parse it is an error.

pkg/isoeditor/stream.go
ori-amizur: So every caller of this function will call it (and handle error from it) before calling `readerForContent`. Something like
```
    start, length, err := boundariesFinder(isoPath, filePath)
    if err != nil {
           return nil, nil, err
    }
   return readerForContent(start, length, base, contentRader)
```

internal/handlers/iso_test.go
```diff
@@ -100,6 +100,15 @@ var _ = Describe("ServeHTTP", func() {
 			)
 		}
 
+		infraenvKargsHandler := func() {
```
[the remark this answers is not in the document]
ori-amizur: Changed to `setInfraenvKargsHandler`

internal/handlers/iso_test.go
```diff
@@ -100,6 +100,15 @@ var _ = Describe("ServeHTTP", func() {
...
+			assistedServer.AppendHandlers(
+				ghttp.CombineHandlers(
+					ghttp.VerifyRequest("GET", fmt.Sprintf(infraEnvPathFormat, imageID)),
+					ghttp.RespondWith(http.StatusOK, "{}", header),
```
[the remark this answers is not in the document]
ori-amizur: OK

pkg/isoeditor/stream_test.go
ori-amizur: Added

pkg/isoeditor/kargs.go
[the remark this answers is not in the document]
carbonin: Okay, I'm fine with this then.

pkg/isoeditor/isoutil.go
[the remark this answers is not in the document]
carbonin: Yeah, I'd rather not change this unless it's really necessary.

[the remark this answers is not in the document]
ori-amizur: Changed

carbonin:
> In new Openshift versions, there is a file that contains the file names, while in old versions the file does not exist. If the file exists, the files to embed the kernel arguments are taken from this config file.

As a separate enhancement we should probably expand our use of this file to use it to determine where to add the rootfs parameter when creating the minimal iso.

This would probably solve some problems (or at least minimize the amount we would need to fix) for the IBM folks doing power and Z integration.
cc @mtarsel @jacobemery

internal/handlers/iso_test.go
```diff
@@ -303,6 +346,82 @@ var _ = Describe("ServeHTTP", func() {
 			expectSuccessfulResponse(resp, []byte("someisocontent"))
 		})
 
+		It("passes non empty kargs", func() {
```
carbonin, line 349: I think a lot of the setup in these tests would be done for you if you put it in the `with no auth` context, right?

Plus it would make sense since you're not specifically testing auth here.

pkg/isoeditor/kargs_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+### BEGIN /etc/grub.d/10_linux ###
+menuentry 'Fedora CoreOS (Live)' --class fedora --class gnu-linux --class gnu --class os {
+	linux /images/pxeboot/vmlinuz mitigations=auto,nosmt coreos.liveiso=fedora-coreos-35.20220103.3.0 ignition.firstboot ignition.platform.id=metal
+	initrd /images/pxeboot/initrd.img /images/ignition.img
+}
+`
+)
+
+var _ = Describe("kargs tests", func() {
```
carbonin, line 62: I think these should be separate `Describe` blocks.

Each `Describe` should list a particular function. So here I'd expect something like:

```go
var _ = Context("with karg helpers", func() {
  const(
    ...
  )

  mockFileReaderCreator := func...
  // more helpers

  Describe("kargsFiles", func() {
    ...
  })

  Describe("kargsEmbedAreaBoundariesFinder", func() {
    ...
  })
})
```

That also gets the constants out of the package scope so they can't accidentally be referenced by non-test code. It also helps with some auto complete engines that might be able to read the scope of identifiers or something.

carbonin (commented):
Just a few test comments. Looks good otherwise though.

pkg/isoeditor/kargs_test.go
ori-amizur: BTW, go does not allow reference from non-test code to test code.
Will change.

internal/handlers/iso_test.go
ori-amizur: With the test `fail kargs infra env query` it is OK, but with the test `passes non empty kargs` I needed to change the `mockImageStream` function.

pkg/isoeditor/kargs_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	mockBoundariesFinderSuccess := func(start, length int64) BoundariesFinder {
+		return mockBoundariesFinderCreator(start, length, nil)
+	}
+
+	mockBoundariesFinderFailure := func() BoundariesFinder {
+		return mockBoundariesFinderCreator(0, 0, errors.New("this is an error"))
+	}
+
+	Context("kargs files", func() {
```
carbonin, line 90: I'm going to reply to your old comment here just so it's not outdated, but TLDR is this should be `Describe("kargsFiles"` because it's grouping tests of that function. It's good to have the function name right in the test.

---------------------------------------------------------------------------------------
In reply to https://github.com/openshift/assisted-image-service/pull/99#discussion_r1005259247:

> BTW, go does not allow reference from non-test code to test code.

This is interesting. I would expect this if the tests were in a different package, but it's surprising to me that it seems to omit `*_test.go` files when compiling main.
That said, if you reference things defined in test files from real ones and run the tests, it works.
To me it's unintuitive and feels like "magic" so I'd like to avoid defining stuff at the top level in tests anyway.

> I think `Describe` is unexpected within `Context` although it is allowed.

I don't think so at all. `Describe` and `Context` are functionally identical. They only differ semantically. Generally I use `Describe` to group tests of a particular function and `Context` when I need shared setup for tests of multiple functions.
If you need shared setup only for the tests of a single function, I'd put it in that function's `Describe` block.

https://github.com/openshift/assisted-image-service/blob/d7fecdd94b334150af7daf53b9fca04722d32582/pkg/isoeditor/isoutil_test.go is a good example where multiple `Describe`s need the same context.

> The text with karg helpers looks unclear to me.

The exact text isn't super important in this case since every test is in the same context. It might as well stay `"kargs tests"`

TLDR is `Describe` is for grouping `It` blocks for the same function and `Context` is for denoting shared setup/teardown.

ori-amizur: > That said, if you reference things defined in test files from real ones and run the tests, it works.

No it doesn't

I usually use `Describe` as the most external function.  That's the most common use in the assisted service.  There are many few external `Context` functions.

If you think it is very significant I will change.

carbonin: >No it doesn't

I wouldn't have asserted something like this if I didn't test it :wink: 
I have a module written that reproduces this behavior, but that said I did also find https://github.com/golang/go/issues/10184#issuecomment-84465873 which explains how it could never cause a _real_ problem.

> If you think it is very significant I will change.

Not _very_ significant, but I wrote this all up mostly to just explain how I think about writing tests. If my descriptions and examples haven't impacted how you want to write these up that's okay. As I said `Context` and `Describe` are functionally equivalent.

The only thing I would very much like to see is the container (`Context` or `Describe`) that is grouping tests for a particular function be named after that function. That way it is very clear what is being tested when a test fails.

ori-amizur: > I wouldn't have asserted something like this if I didn't test it wink
I have a module written that reproduces this behavior, but that said I did also find https://github.com/golang/go/issues/10184#issuecomment-84465873 which explains how it could never cause a real problem.

Well that's strange because I tested it too.
In the same package, I tried to reference in non-test file a `const` that is defined in test file.  The symbol was invisible.
Specifically tried on the `const` you made the initial comment on.

pkg/isoeditor/kargs_test.go
ori-amizur: Switched `Describe` and `Context`

carbonin (approved):

PR #105, opened 2022-12-02
Validate the volume ID for downloaded ISOs

carbonin:
## Description

Before this change a user could add a non-rhcos/fcos ISO to our image list and we would fail in unpredictable ways with unhelpful error messages. This commit explicitly validates the volume ID on the downloaded ISO to ensure we're dealing with something we support.

With this change the image service fails with a clear error message when configured with an unsupported ISO:
```
{"file":"/go/src/github.com/openshift/origin/pkg/imagestore/imagestore.go:210","func":"github.com/openshift/assisted-image-service/pkg/imagestore.(*rhcosStore).Populate.func1","level":"error","msg":"failed to validate /data/rhcos-full-iso-4.8-48.84.202109241901-0-x86_64.iso: ISO volume identifier (Fedora-S-dvd-x86_64-37) is invalid","time":"2022-12-02T15:00:52Z"}
{"file":"/go/src/github.com/openshift/origin/main.go:74","func":"main.main.func1","level":"fatal","msg":"Failed to populate image store: failed to validate /data/rhcos-full-iso-4.8-48.84.202109241901-0-x86_64.iso: ISO volume identifier (Fedora-S-dvd-x86_64-37) is invalid\n","time":"2022-12-02T15:00:52Z"}
```

This change also removes the invalid full ISO. Before, the full ISO would be cached and wouldn't be removed upon failure.

## How was this code tested?

```
$ export OS_IMAGES='[{"openshift_version": "4.8","cpu_architecture": "x86_64","url": "https://download.fedoraproject.org/pub/fedora/linux/releases/37/Server/x86_64/iso/Fedora-Server-dvd-x86_64-37-1.7.iso","version": "48.84.202109241901-0"}]'

$ IMAGE=quay.io/carbonin/assisted-image-service:latest make build run
```

On the `main` branch this results in the following fatal logs:
```
{"file":"/go/src/github.com/openshift/origin/pkg/imagestore/imagestore.go:210","func":"github.com/openshift/assisted-image-service/pkg/imagestore.(*rhcosStore).Populate","level":"info","msg":"Creating minimal iso for 4.8-48.84.202109241901-0-x86_64","time":"2022-12-02T15:05:42Z"}
{"file":"/go/src/github.com/openshift/origin/main.go:74","func":"main.main.func1","level":"fatal","msg":"Failed to populate image store: failed to create minimal iso template for version map[cpu_architecture:x86_64 openshift_version:4.8 url:https://download.fedoraproject.org/pub/fedora/linux/releases/37/Server/x86_64/iso/Fedora-Server-dvd-x86_64-37-1.7.iso version:48.84.202109241901-0]: remove /data/isoutil2549557190/images/pxeboot/rootfs.img: no such file or directory\n","time":"2022-12-02T15:05:51Z"}
```

With this patch the logs are more descriptive (show that the ISO isn't one we expect).

## Assignees

/cc @CrystalChun 
/cc @danielerez 

## Links

Resolves https://issues.redhat.com/browse/MGMT-12740

## Checklist

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/renameio"
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
@@ -164,6 +165,19 @@ func (s *rhcosStore) downloadURLToFile(url string, path string) error {
 	return nil
 }
 
+func validateISOID(path string) error {
+	volumeID, err := isoeditor.VolumeIdentifier(path)
+	if err != nil {
+		return err
+	}
+
+	if !strings.HasPrefix(volumeID, "rhcos-") && !strings.HasPrefix(volumeID, "fedora-coreos-") {
+		return fmt.Errorf("ISO volume identifier (%s) is invalid", volumeID)
+	}
+
+	return nil
+}
+
 func (s *rhcosStore) Populate(ctx context.Context) error {
 	if err := s.cleanDataDir(); err != nil {
 		return err
@@ -187,8 +201,15 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 				if err != nil {
 					return fmt.Errorf("failed to download %s: %v", url, err)
 				}
-
 				log.Infof("Finished downloading for %s-%s (%s)", openshiftVersion, arch, imageVersion)
+				if err := validateISOID(fullPath); err != nil {
+					message := fmt.Sprintf("failed to validate %s: %v", fullPath, err)
+					if err = os.Remove(fullPath); err != nil {
+						log.WithError(err).Errorf("failed to remove invalid ISO %s", fullPath)
+					}
+					log.Error(message)
+					return fmt.Errorf(message)
+				}
 			}
 
 			return nil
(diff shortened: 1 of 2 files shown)


CrystalChun:
Good catch! I'm not as familiar with this so I have a few questions:
1. What would the state of the image service be if:
    a. Only one of the ISOs is unsupported?
    b. All of the ISOs specified are unsupported?
2. Would it be a good idea to include the `openshift_version` and `cpu_architecture` in the error message since the downloaded file might be easily mappable to what they provided in `OS_IMAGES`?
3. How would a user fix this issue? Do they edit the `OS_IMAGES` environment variable and restart the pod?

carbonin:
> 1. What would the state of the image service be if:
>  a. Only one of the ISOs is unsupported?
>  b. All of the ISOs specified are unsupported?

In both cases the service exits on the first unsupported one it finds. I did this intentionally because I didn't want the first error to be when the user goes to download something.

The issue is with the deployment (env vars in this case) so it should fail to deploy.

> 2. Would it be a good idea to include the openshift_version and cpu_architecture in the error message since the downloaded file might be easily mappable to what they provided in OS_IMAGES?

The filename in the logs is based directly on all of this so I thought it would be easy enough to understand which was the problem `failed to validate /data/rhcos-full-iso-4.8-48.84.202109241901-0-x86_64.iso`

> 3. How would a user fix this issue? Do they edit the OS_IMAGES environment variable and restart the pod?

Yeah, they would need to change the URL in `OS_IMAGES` to point to something supported.

CrystalChun (approved):
Gotcha thanks for answering my questions!
/lgtm

PR #109, opened 2022-12-20
Support ppc64le and s390x

jacobemery:
## Description

@carbonin and I worked on this together to enable ppc64le (Power) and s390x (Z) architectures in the assisted-image-service. This is related to the [PR](https://github.com/openshift/assisted-service/pull/4796) in the assisted-service to enable these architectures related to this multi-arch [epic](https://issues.redhat.com/browse/MULTIARCH-2889).

We're working around some limitations with s390x as it doesn't use grub, which also means we can't extract a minimal iso from it. The CoreOS team is [working on](https://issues.redhat.com/browse/COS-1924) enabling minimal iso extraction for s390x at the moment. If/when they get it working we can revisit some of these changes, but for now we're just going to skip minimal iso extraction for s390x.

Neither architecture has an isolinux.cfg file, so we had to make some changes to avoid errors when trying to edit it. 

Also made some changes to the integration tests to take these two architectures into account and avoid errors.

## How was this code tested?

Integration tests validate that we can download the ISOs of these architectures.

In the assisted-service [PR](https://github.com/openshift/assisted-service/pull/4796), we will do additional testing.

## Assignees

/cc @mkowalski 
/cc @danielerez 

## Links

[Epic](https://issues.redhat.com/browse/MULTIARCH-2889) for enabling P/Z in Assisted Installer
[PR](https://github.com/openshift/assisted-service/pull/4796) for assisted-service
[Ticket](https://issues.redhat.com/browse/COS-1924) for CoreOS to enable minimal ISO extraction 

## Checklist

- [X] Not yet - This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Unsure - Does this change include unit tests (note that code changes require unit tests)

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -241,7 +241,7 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 				return fmt.Errorf("failed to build rootfs URL: %v", err)
 			}
 
-			err = s.isoEditor.CreateMinimalISOTemplate(fullPath, rootfsURL, minimalPath)
+			err = s.isoEditor.CreateMinimalISOTemplate(fullPath, rootfsURL, arch, minimalPath)
 			if err != nil {
 				return fmt.Errorf("failed to create minimal iso template for version %s: %v", imageInfo, err)
 			}
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -226,6 +226,11 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 		imageVersion := imageInfo["version"]
 		arch := imageInfo["cpu_architecture"]
 
+		// Don't attempt to create a minimal ISO for s390x because there's no easy way to edit the kernel parameters
+		// This means that the rootfs URL can't be added which makes it impossible for us to create a minimal ISO
+		if arch == "s390x" {
+			continue
+		}
 		minimalPath := filepath.Join(s.dataDir, isoFileName(ImageTypeMinimal, openshiftVersion, imageVersion, arch))
 		if _, err := os.Stat(minimalPath); os.IsNotExist(err) {
 			log.Infof("Creating minimal iso for %s-%s-%s", openshiftVersion, imageVersion, arch)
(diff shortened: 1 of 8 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -226,6 +226,9 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 		imageVersion := imageInfo["version"]
 		arch := imageInfo["cpu_architecture"]
 
+		if arch == "s390x" {
+			continue
```
danielerez, line 230: Do you think it's worth adding a comment here about the reason we skip this arch? I.e. maybe a reference to the issue preventing the extracting iso.

pkg/isoeditor/rhcos.go
```diff
@@ -87,8 +93,8 @@ func embedInitrdPlaceholders(extractDir string) error {
 	return nil
 }
 
-func fixTemplateConfigs(rootFSURL, extractDir string) error {
-	availableGrubPaths := []string{"EFI/redhat/grub.cfg", "EFI/fedora/grub.cfg"}
+func fixGrubConfig(rootFSURL, extractDir string) error {
+	availableGrubPaths := []string{"EFI/redhat/grub.cfg", "EFI/fedora/grub.cfg", "boot/grub/grub.cfg"}
```
danielerez, line 97: So iiuc, no EFI folder in ppc/s390x images?

danielerez (commented):
Looks good. Just a couple of minor comments.

pkg/isoeditor/rhcos.go
carbonin: There's no `grub.cfg` in x390x at all, but it's just in a different place for ppc64le

pkg/imagestore/imagestore.go
carbonin: Sure, I'll add a comment and explain it more thoroughly in the commit message.

pkg/isoeditor/rhcos.go
jacobemery: Yes, ppc64le and s390x don't use EFI

PR #114, opened 2023-02-14
validateISOID: allow SCOS images

vrutkovs:
## Description

Allow images with "scos-" volume ID. This is required to setup OKD SCOS via Assisted Installer

## How was this code tested?

Manual run

## Assignees

/cc @carbonin 
/cc @

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -72,6 +72,12 @@ var (
 			"url":               "https://mirror.openshift.com/pub/openshift-v4/ppc64le/dependencies/rhcos/4.11/4.11.9/rhcos-4.11.9-ppc64le-live.ppc64le.iso",
 			"version":           "ppc64le-latest",
 		},
+		{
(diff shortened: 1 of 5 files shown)


vrutkovs:
Added unit tests, but since we don't have SCOS images pushed to official S3 bucket I can add integration test later. Lets

/hold

it for now

vrutkovs:
/hold cancel

SCOS builds are not public, added integration test

carbonin:
> SCOS builds are not public, added integration test

What does this mean for us? I mean, you added a test so they are downloadable, but they're not public in that they don't get published the same way?

vrutkovs:
>What does this mean for us? I mean, you added a test so they are downloadable, but they're not public in that they don't get published the same way?

Oh, sorry, there _were_ not public when I wrote that comment and I assumed it would take a while to get them published. Luckily it was easy and SCOS builds are downloadable freely

carbonin (approved):

PR #119, opened 2023-04-07
adds ability to set the log level

mhrivnak:
## Description
The log level can now be set via environment variable, enabling developers and operators to see debug messages.

## How was this code tested?
Hand testing and the existing suite of integration tests.

## Assignees

/cc @carbonin 

## Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,6 +36,7 @@ var Options struct {
 	AllowedDomains        string `envconfig:"ALLOWED_DOMAINS"`
 	InsecureSkipVerify    bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
 	ImageServiceBaseURL   string `envconfig:"IMAGE_SERVICE_BASE_URL"`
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/assisted-image-service/pkg/servers"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
```
carbonin, line 19: Can we deduplicate this in any sane way? Maybe its better to not use a global logger or would that be too much change for this PR?

mhrivnak: Ah, I think that's just an oversight on my part, not noticing how the IDE added another import for it. Should be easy to get rid of the extra import.

carbonin (approved):

PR #120, opened 2023-04-12
MGMT-12743: adds support for shorter URLs

mhrivnak:
## Description

Implements the [Short Image URL
enhancement](https://github.com/openshift/assisted-service/blob/master/docs/enhancements/short-url-enhancement.md).

Quoting the enhancement summary: "The assisted-image-service should be enhanced to serve discovery ISOs from URLs that are less than 255 characters and don't have query parameters. This will enable servers from multiple vendors to directly boot from an assisted-image-service URL using the baseboard management controller."

## How was this code tested?
by hand and with new automated tests

## Assignees

/assign @carbonin 

## Checklist

### DIFF
--- a/internal/handlers/short.go
+++ b/internal/handlers/short.go
@@ -1,0 +1,91 @@
+package handlers
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"regexp"
+
+	"github.com/go-chi/chi/v5"
+)
+
+var jwtPayloadRegexp = regexp.MustCompile(`^.+\.(.+)\..+`)
+
+type payload struct {
+	Sub        string `json:"sub"`          // used by OCM tokens
+	InfraEnvID string `json:"infra_env_id"` // used by local auth tokens
+}
+
+// parseShortURL parses short-style URLs, where URL path segments are used to
+// hierarchically identify the desired resource. The URL ordering and structure
+// is defined on the router.
+func parseShortURL(r *http.Request) (*imageDownloadParams, int, error) {
+	imageID := chi.URLParam(r, "image_id")
+	token := chi.URLParam(r, "token")
+	if token == "" {
+		// try the "api_key" parameter name, which is used at the /byapikey/ path.
+		token = chi.URLParam(r, "api_key")
+	}
+	version := chi.URLParam(r, "version")
+	arch := chi.URLParam(r, "arch")
+	filename := chi.URLParam(r, "filename")
+
+	var err error
+	// the URL can have either the token or the image_id
+	if imageID == "" {
+		imageID, err = idFromJWT(token)
+		if err != nil {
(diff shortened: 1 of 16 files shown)


internal/handlers/short.go
```diff
@@ -0,0 +1,87 @@
...
+	case "full.iso":
+		params.imageType = "full-iso"
+	default:
+		return nil, http.StatusNotFound, fmt.Errorf("unrecognized file name %s", filename)
+	}
+
+	return &params, 0, nil
+}
+
+func idFromJWT(jwt string) (string, error) {
```
carbonin, line 61: Maybe add a comment here that we acknowledge we're parsing and using the contents of the token without verifying it because that will be done by assisted later.

Also you can use the jwt library to do this for you (as we do in [assisted](https://github.com/openshift/assisted-service/blob/41251e4325c8faa041b139e2ee818de18969b5e5/internal/gencrypto/token.go#L93-L106)), but I'm just as happy to not add an additional dependency since the regex is rather simple and incredibly unlikely to change.

internal/handlers/short_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+
+	"github.com/go-chi/chi/v5"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Parse short URLs", func() {
```
carbonin, line 15: I'd prefer to use `Describe` to identify the function under test and split different groups into different `Describe` or `Context` blocks.

So this would be `Describe("idFromJWT" ...` containing tests like:
`succeeds with a JWT with an infra env ID`
`fails with an invalid JWT`
`fails without a valid claim`
etc.

The test name along with the containing contexts and describe blocks should tell you the function, scenario, and expected outcome in something approximating English.

internal/handlers/short_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("parse JWT with no ID", func() {
+		_, err := idFromJWT(tokenNoID)
+
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("200 if infra_env_id present in token", func() {
```
carbonin, line 73: This test and the following ones should be in a `Describe("parseShortURL"...` block

carbonin (commented):
Just a few comments about the format of the new tests.

Otherwise the content looks good. Do you have a matching PR for assisted to start handing out these URLs or is that still WIP?

internal/handlers/short.go
mhrivnak: I'll add a comment. And I agree that it's not worth pulling in a whole library to replace a few simple lines of code.

carbonin (approved):

PR #123, opened 2023-04-24
Allow kernel downloads for s390x architecture

carbonin:
## Description

The s390x ISO kernel is at `/images/pxeboot/kernel.img` instead of `/images/pxeboot/vmlinuz`. This PR accounts for that difference.

It also updates the template to reference existing URLs for the 4.13 images since I noticed these were wrong when trying to find an image to test with.

## How was this code tested?
Ran the new image locally and verified that the kernel downloads correctly.

## Assignees

/cc @CrystalChun 
/cc @danielerez 
/cc @jaypoulz 
/cc @jacobemery 

## Links

https://issues.redhat.com/browse/MGMT-13716

## Checklist

### DIFF
--- a/deploy/template.yaml
+++ b/deploy/template.yaml
@@ -83,25 +83,25 @@ parameters:
       {
         "openshift_version": "4.13",
         "cpu_architecture": "x86_64",
-        "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.13/rhcos-4.13.0-ec.3-x86_64-live.x86_64.iso",
+        "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/4.13.0-rc.2/rhcos-4.13.0-rc.2-x86_64-live.x86_64.iso",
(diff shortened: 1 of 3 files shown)


deploy/template.yaml
```diff
@@ -83,25 +83,25 @@ parameters:
       {
         "openshift_version": "4.13",
         "cpu_architecture": "x86_64",
-        "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.13/rhcos-4.13.0-ec.3-x86_64-live.x86_64.iso",
+        "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/4.13.0-rc.2/rhcos-4.13.0-rc.2-x86_64-live.x86_64.iso",
```
filanov, line 86: Do we need to update anything in the assisted service?

filanov (approved):

deploy/template.yaml
carbonin: Nope, they were updated in https://github.com/openshift/assisted-service/pull/5115

PR #137, opened 2023-08-28
AGENT-702: Generate minimal ISO for agent installer

pawanpinjarkar:
## Description

Generate minimal ISO for agent installer similar to the assisted installer by removing the rootfs image file. There are minor differences compared to the existing method for assisted installer such as 
- The agent installer will provide the value for volume ID at the time of invocation
- When `CreateMinimalISO()` is invoked from the agent installer, the agent installer provides the location of extracted files via `extractDir` so `CreateMinimalISO` does not need to extract the ISO whereas the assisted installer will first extract the ISO in `CreateMinimalISOTemplate()` and then call `CreateMinimalISO()`

## Assignees

/cc @
/cc @

## Links

https://issues.redhat.com/browse/AGENT-702

## Checklist

### DIFF
--- a/pkg/isoeditor/rhcos.go
+++ b/pkg/isoeditor/rhcos.go
@@ -28,6 +28,36 @@ func NewEditor(dataDir string) Editor {
 	return &rhcosEditor{workDir: dataDir}
 }
 
+// CreateMinimalISO Creates the minimal iso by removing the rootfs and adding the url
+func CreateMinimalISO(extractDir, volumeID, rootFSURL, arch, minimalISOPath string) error {
+	if err := os.Remove(filepath.Join(extractDir, "images/pxeboot/rootfs.img")); err != nil {
+		return err
+	}
+
+	if err := embedInitrdPlaceholders(extractDir); err != nil {
+		log.WithError(err).Warnf("Failed to embed initrd placeholders")
+		return err
+	}
+
+	if err := fixGrubConfig(rootFSURL, extractDir); err != nil {
+		log.WithError(err).Warnf("Failed to edit grub config")
+		return err
+	}
+
+	// ignore isolinux.cfg for ppc64le because it doesn't exist
+	if arch != "ppc64le" {
+		if err := fixIsolinuxConfig(rootFSURL, extractDir); err != nil {
+			log.WithError(err).Warnf("Failed to edit isolinux config")
+			return err
+		}
+	}
+
+	if err := Create(minimalISOPath, extractDir, volumeID); err != nil {
+		return err
+	}
+	return nil
+}
+
 // CreateMinimalISOTemplate Creates the template minimal iso by removing the rootfs and adding the url
 func (e *rhcosEditor) CreateMinimalISOTemplate(fullISOPath, rootFSURL, arch, minimalISOPath string) error {
 	extractDir, err := os.MkdirTemp(e.workDir, "isoutil")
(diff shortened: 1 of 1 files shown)


pkg/isoeditor/rhcos.go
```diff
@@ -72,8 +70,40 @@ func (e *rhcosEditor) CreateMinimalISOTemplate(fullISOPath, rootFSURL, arch, min
...
+	if err := Create(agentIsoFile, extractDir, volumeID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func embedInitrdPlaceholders(extractDir, rdImagePath string) error {
+	if rdImagePath == "" {
+		rdImagePath = ramDiskImagePath
```
carbonin, line 104: Instead of this pass `ramDiskImagePath` from the callers that need it.
Don't mess around with empty/default parameters.

carbonin (commented):
I think this could be better described as "

carbonin:
Oh I see, you also don't want to extract the image.

carbonin (changes requested):
I think there's a better way to refactor this so more code is shared and we do less parameter defaulting.

Maybe create a function like `createMinimalISOFromDir` that both agent and template functions could use.

It could take the initrd path (if that's necessary) and the volumeID as parameters.

pkg/isoeditor/rhcos.go
pawanpinjarkar: Restructured.  I think the agent installer doesn't care what the value of ramDiskImagePath is so we can use the default one here and not ask agent installer to pass

pkg/isoeditor/rhcos.go
```diff
@@ -28,44 +28,63 @@ func NewEditor(dataDir string) Editor {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateMinimalISO Creates the minimal iso by removing the rootfs and adding the url for agent installer
+func CreateMinimalISO(minimalISOPath, extractDir, volumeID, rootFSURL, arch string) error {
+	err := createMinimalISOFromDir(minimalISOPath, extractDir, volumeID, rootFSURL, arch)
```
pawanpinjarkar, line 86: This can further be changed to `CreateMinimalISOFromDir` i.e. exported and agent installer can call it.

pkg/isoeditor/rhcos.go
```diff
@@ -28,44 +28,63 @@ func NewEditor(dataDir string) Editor {
...
+		return err
+	}
+
 	volumeID, err := VolumeIdentifier(fullISOPath)
 	if err != nil {
 		return err
 	}
 
-	if err = Create(minimalISOPath, extractDir, volumeID); err != nil {
+	err = createMinimalISOFromDir(minimalISOPath, extractDir, volumeID, rootFSURL, arch)
```
zaneb, line 76: nit: the change of argument order here, where `minimalISOPath` went from the end to the beginning (in place of `fullISOPath`) makes for an API that is prone to mistakes/misunderstandings.

pkg/isoeditor/rhcos.go
zaneb: TBH defining another function just to have a different name exported externally vs internally seems a bit unnecessary.
Perhaps you could rename `createMinimalISOFromDir` to e.g. `AssembleMinimalISO` or somthing.

pkg/isoeditor/rhcos.go
```diff
@@ -28,44 +28,63 @@ func NewEditor(dataDir string) Editor {
...
-		return err
-	}
-
-	if err = Extract(fullISOPath, extractDir); err != nil {
-		return err
-	}
-
-	if err = os.Remove(filepath.Join(extractDir, "images/pxeboot/rootfs.img")); err != nil {
+func createMinimalISOFromDir(minimalISOPath, extractDir, volumeID, rootFSURL, arch string) error {
+	if err := os.Remove(filepath.Join(extractDir, "images/pxeboot/rootfs.img")); err != nil {
```
zaneb, line 32: Arguably we could also ignore NotFound errors here, now that the caller is passing in a directory rather than the contents being extracted inline.

pkg/isoeditor/rhcos.go
pawanpinjarkar: Done

pkg/isoeditor/rhcos.go
pawanpinjarkar: Ok, so renamed the func. The original `CreateMinimalISOTemplate()` now calls `CreateMinimalISO()`. The actual minimal ISO is "created" in `CreateMinimalISO()` and the same will be called by agent installer

pkg/isoeditor/rhcos.go
pawanpinjarkar: Agree, but then instead of ignoring, may be it doesn't hurt to have this check.

carbonin: I would rather have the check in case coreos changes the path on us.
We would really need to know that the rootfs isn't getting deleted.

carbonin (approved):

PR #139, opened 2023-09-06
[WIP] MGMT-13692: Use the ignition file info

carbonin:
## Description

If the ignition info file (/coreos/igninfo.json) is present in the ISO
use it to determine where the ignition content should be embedded
instead of assuming it should always go in "/images/ignition.img"

If an error is encountered when reading the ignition info, fall back to
the previous ignition path. This allows the image service to seamlessly
support older versions without having to explicitly include logic about
which version is being used in every call.

## How was this code tested?

Added a failing test in the integration tests (s390x pre-release).
This should be fixed before merging.

I'm not sure if it's worth adding separate unit-tests for with or without this file.
For now maybe we should pin a single integration test to a version without it? I'm not sure.

## Assignees

/cc @filanov 
/cc @ori-amizur 

## Links

Resolves https://issues.redhat.com/browse/MGMT-13692

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -31,22 +31,52 @@ import (
 var (
 	versions = []map[string]string{
 		{
-			"openshift_version": "pre-release",
+			"openshift_version": "pre-release-arm",
 			"cpu_architecture":  "arm64",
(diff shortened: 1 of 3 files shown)


jhernand:
@carbonin I have been asked to continue your work on this, see #156. I removed your commit to modify the integration tests and added a commit to add unit tests instead.

PR #145, opened 2023-09-26
MGMT-15822: change base image to stream9

osherdp:
## Description

Now that D/S building system fully supports RHEL-9 base images (and allows easy installation of nmstate and other libs that were previously unsupported) we can move U/S to stream9 and D/S to ubi9-minimal.

It doesn't seem possible to easily use ubi9-minimal U/S for the assisted-service, because installation of nmstate packages either requires entitlements running on a RHEL node, or alternatively an access to RedHat's internal network (which can be a major problem for U/S development). For a better consistency we're sticking with the same for the agent, installer, etc.

## How was this code tested?

CI will do the tests.

## Assignees

TBD

## Links

https://issues.redhat.com/browse/MGMT-15822

## Checklist

### DIFF
--- a/Dockerfile.image-service
+++ b/Dockerfile.image-service
@@ -6,7 +6,7 @@ RUN go mod download
 COPY . .
 RUN CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /assisted-image-service main.go
 
-FROM quay.io/centos/centos:stream8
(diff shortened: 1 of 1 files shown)


osherdp:
/hold
waiting for tests' completion

adriengentil (approved):

PR #154, opened 2023-10-27
OCPBUGS-22427: Use 'igninfo.json' to find ignition location

jhernand:
## Description

Since version 0.17.0 of the CoreOS installer there is a new `/coreos/igninfo.json` file that indicates the location of the ignition configuration inside the ISO. For example, for version 4.14 the S390 live ISO contains this:

```json
{
  "file": "images/cdboot.img",
  "length": 262144,
  "offset": 66497660
}
```

But we currently ignore this file and expect the ignition configuration always in `/images/ignition.img`, and therefore we fail for these S390 live ISO.

This patch changes the service so that it always looks for the `igninfo.json` file. If it exists it will use the location it indicates, if it doesn't exist it will use the default location.

## How was this code tested?

Tested with the included unit tests.

## Assignees

/cc @gamli75 

## Links

Related: https://issues.redhat.com/browse/OCPBUGS-22427
Related: https://github.com/coreos/coreos-installer/releases/tag/v0.17.0

## Checklist

### DIFF
--- a/pkg/isoeditor/stream.go
+++ b/pkg/isoeditor/stream.go
@@ -2,21 +2,39 @@ package isoeditor
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 
 	"github.com/openshift/assisted-image-service/pkg/overlay"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 )
 
-const ignitionImagePath = "/images/ignition.img"
+const (
+	ignitionInfoPath         = "/coreos/igninfo.json"
+	defaultIgnitionImagePath = "/images/ignition.img"
+)
 
 type ImageReader = overlay.OverlayReader
 
 type BoundariesFinder func(filePath, isoPath string) (int64, int64, error)
 
 type StreamGeneratorFunc func(isoPath string, ignitionContent *IgnitionContent, ramdiskContent, kargs []byte) (ImageReader, error)
+
+// IgnInfo us used to read and write the content of the '/coreos/igninfo.json' file that indicates
+// the location of the ignition configuration inside the ISO.
+type IgnInfo struct {
+	// File is the absolute path of the file containing the ignition configuration.
+	File string `json:"file"`
+
+	// Offset is the offset of the ignition configuration file within the file.
+	Offset int64 `json:"offset,omitempty"`
+
+	// Length is the length of the ignition configuration.
+	Length int64 `json:"length,omitempty"`
+}
 
 func NewRHCOSStreamReader(isoPath string, ignitionContent *IgnitionContent, ramdiskContent []byte, kargs []byte) (ImageReader, error) {
 	isoReader, err := os.Open(isoPath)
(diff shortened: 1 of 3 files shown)


pkg/isoeditor/stream.go
```diff
@@ -29,7 +47,69 @@ func NewRHCOSStreamReader(isoPath string, ignitionContent *IgnitionContent, ramd
...
+	//
+	//	{
+	//		"file": "images/cdboot.img",
+	//		"offset": 66497660,
+	//		"length": 262144
+	//	}
+	//
+	// If it exists we need to read it, otherwise we can use the '/images/ignition.img' location
+	// used in previous versions.
+	var ignInfo IgnInfo
```
filanov, line 62: Suggesting so separate the code that get the `ignInfo` into a different function `NewRHCOSStreamReader` is already long enough

pkg/isoeditor/stream.go
```diff
@@ -29,7 +47,69 @@ func NewRHCOSStreamReader(isoPath string, ignitionContent *IgnitionContent, ramd
...
+	ignInfoBytes, err := ReadFileFromISO(isoPath, ignitionInfoPath)
```
tsorya, line 63: Can't we validate exact error or we always want to retry?

pkg/isoeditor/stream.go
```diff
@@ -29,7 +47,69 @@ func NewRHCOSStreamReader(isoPath string, ignitionContent *IgnitionContent, ramd
...
+	if err != nil {
+		log.WithError(err).WithFields(log.Fields{
```
nmagnezi, line 65: I'd say this should not be logged as an error, as we know this is done for backward compatibility.
Maybe change it to warning?

pkg/isoeditor/stream.go
nmagnezi: +1

gamli75:
@jhernand it looks like Nick started to work on it, please review https://github.com/openshift/assisted-image-service/pull/139

pkg/isoeditor/stream.go
jhernand: Is not logged as an error, but as `INFO`. The `WithError` call only means "add this error as a parameter". The resulting message is something like this:

```
time="2023-10-30T08:25:40+01:00" level=info msg="Failed to read ignition information file, will assume that it doesn't exist and use the default ignition location" default=/images/ignition.img error="could not read directory entries for /coreos" file=/coreos/igninfo.json iso=/tmp/1408693960test.iso
```

I can change it to warning if you prefer that.

pkg/isoeditor/stream.go
jhernand: Unfortunatelly the diskfs library doesn't return `os.ErrNotExist`, it returns a generic string error with this message if the file doesn't exist:

```
could not read directory entries for /coreos
```

I could check that string, but I thought it is fragile and not worth. Trying always with the default doesn't hurt.

jhernand:
> @jhernand it looks like Nick started to work on it, please review #139

That change by Nick looks file to me. Should I close this in favor of that?

pkg/isoeditor/stream.go
jhernand: I will do that if we decide to use this instead of Nick's #139.

filanov:
> > @jhernand it looks like Nick started to work on it, please review #139
> 
> That change by Nick looks file to me. Should I close this in favor of that?

His code looks a bit simpler, so maybe it be worth using it.

jhernand:
> His code looks a bit simpler, so maybe it be worth using it.

No problem, I am closing this then.

PR #163, opened 2023-11-16
MGMT-16239: Use ignition info - ACM 2.9

jhernand:
## Description

This is a backport of [MGMT-13692](https://issues.redhat.com/browse/MGMT-13692) and #156 for ACM 2.9.

Note that this is mostly a copy of #139. The differences is that there are no changes in the integration tests, and a new unit test was added to verify the use of `igninfo.json` in S390 ISOs. Below is the description originally written by @carbonin:

> If the ignition info file (/coreos/igninfo.json) is present in the ISO
> use it to determine where the ignition content should be embedded
> instead of assuming it should always go in "/images/ignition.img"
>
> If an error is encountered when reading the ignition info, fall back to
> the previous ignition path. This allows the image service to seamlessly
> support older versions without having to explicitly include logic about
> which version is being used in every call.

## How was this code tested?

Added a unit test that creates a ISO that simulates the `igninfo.json`  file of the S390 architecture.

## Assignees

/cc @filanov
/cc @ori-amizur

## Links

Related: https://issues.redhat.com/browse/MGMT-16239
Related: https://issues.redhat.com/browse/MGMT-13692
Related: https://github.com/openshift/assisted-image-service/pull/156

## Checklist

### DIFF
--- a/pkg/isoeditor/isoeditor_suite_test.go
+++ b/pkg/isoeditor/isoeditor_suite_test.go
@@ -1,6 +1,8 @@
 package isoeditor
 
 import (
+	"crypto/rand"
+	"encoding/json"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


jhernand:
@oourfali I had to force push this again in order to re-run the tests, `/retest` wasn't working. It will need a `/lgtm` again.

filanov (approved):

PR #168, opened 2023-12-20
Respect `HTTPS_CA_FILE` in the image store client

omertuc:
## Description
`HTTPS_CA_FILE` was only trusted by the client of the
image-service->assisted-service communication, not the client of the
image-service->image-store communication. This means that users couldn't
add certs to trust an image store that uses a custom CA

This commit makes it so that both clients will use the certs specified
in `HTTPS_CA_FILE`

## How was this code tested?

Untested, draft

## Assignees

/cc @filanov
/cc @carbonin

## Links

Partially solves MGMT-14070, further changes required in infrastructure
operator agentserviceconfig controller to accept certs that will be added to
the file that it passes to the image service at `HTTPS_CA_FILE`

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ skipper make test
 - `DATA_DIR` - Path at which to store downloaded RHCOS images.
 - `HTTPS_CERT_FILE` - tls cert file path
 - `HTTPS_KEY_FILE` - tls key file path
+- `HTTPS_CA_FILE` - A file containing CA certificates the image service should trust when connecting to the Assisted Service and the image store
 - `HTTP_LISTEN_PORT` - When set, plain http listener is started on that port
 - `IMAGE_SERVICE_BASE_URL` - the base URL to use to query the image service
 - `LISTEN_PORT` - Image Service listen port
--- a/internal/handlers/assisted_service_client.go
+++ b/internal/handlers/assisted_service_client.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


carbonin:
The CM the provides this cert is populated using a special annotation that indicates to openshift to add the in-cluster CA certs to the CM.

I'm not sure we can edit that CM with arbitrary additional certs. Worth checking out before pushing in this direction.

refs: 

Configmap being created in agent service config controller: https://github.com/openshift/assisted-service/blob/bca1fda119ab24f1274adfc5d9673a1e1d1b9d25/internal/controller/controllers/agentserviceconfig_controller.go#L987

Docs on this config map annotation feature: https://docs.openshift.com/container-platform/4.14/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate
Note specifically:
> After adding this annotation to a config map all existing data in it is deleted.

omertuc:
>I'm not sure we can edit that CM with arbitrary additional certs.

Right. infra operator would have to create a new configmap that merges the user given certs and the certs from that uneditable cm

carbonin:
Would it not be easier to leave this the way it is and create a new configmap just for the user certs and use the current one for querying assisted and the new one for querying the images?

There's never going to be overlap between the certs we need to trust in these cases, right?

omertuc:
I don't see how that would necessarily easier, but sounds reasonable, I don't have a preference either way

filanov:
+1 for the separation.

carbonin:
Implemented in https://github.com/openshift/assisted-image-service/pull/172

PR #172, opened 2024-01-15
MGMT-16507: Enable image service to use additional ca certificates

paul-maidment:
[MGMT-16507](https://issues.redhat.com//browse/MGMT-16507): Enable image service to use additional ca certificates

The infrastructure operator will make available a file at ASSISTED_SERVICE_API_TRUSTED_CA_FILE

This file will contain certificates that are to be used for certificate authority during image fetch
The purpose of this PR is to use those certificates for the fetching of images if they have been set up.

Additionally, we needed to make the  the purpose of HTTPS_CA_FILE clearer, so we added a new variable OS_IMAGE_DOWNLOAD_TRUSTED_CA_FILE
The value of OS_IMAGE_DOWNLOAD_TRUSTED_CA_FILE will default to the value held in HTTPS_CA_FILE unless overridden
The idea is to retire HTTPS_CA_FILE as it is not very clear what it's for.

OSImageDownloadTrustedCAFile is a path to a CA file that will be trusted when fetching OS Images
AssistedServiceApiTrustedCAFile is a path to a CA file that will be trusted for TLS connections to the Assisted Service API

## How was this code tested?
This code has coverage in a unit test that ensures we make a call to an HTTPS test server that will require us to use a CA.
Thas additionally been tested manually by deploying the infrastructure operator from a custom bundle.

## Assignees
/cc @carbonin 

## Links
https://github.com/openshift/assisted-service/pull/5884)

## Checklist

There is a task in the epic to handle user specific documentation

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,7 +27,10 @@ var Options struct {
 	DataDir               string `envconfig:"DATA_DIR"`
 	HTTPSKeyFile          string `envconfig:"HTTPS_KEY_FILE"`
 	HTTPSCertFile         string `envconfig:"HTTPS_CERT_FILE"`
-	HTTPSCAFile           string `envconfig:"HTTPS_CA_FILE"`
+
+	// Deprecated - use ASSISTED_SERVICE_API_TRUSTED_CA_FILE instead
+	HTTPSCAFile string `envconfig:"HTTPS_CA_FILE"`
+
 	ListenPort            string `envconfig:"LISTEN_PORT" default:"8080"`
 	HTTPListenPort        string `envconfig:"HTTP_LISTEN_PORT"`
 	MaxConcurrentRequests int64  `envconfig:"MAX_CONCURRENT_REQUESTS" default:"400"`
@@ -37,9 +40,21 @@ var Options struct {
 	InsecureSkipVerify    bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
 	ImageServiceBaseURL   string `envconfig:"IMAGE_SERVICE_BASE_URL"`
 	LogLevel              string `envconfig:"LOGLEVEL" default:"info"`
+
+	// This is a path to a CA file that will be trusted when fetching OS Images
+	// intended for scenarios where the OS images are served from a service that uses a custom CA
+	OSImageDownloadTrustedCAFile string `envconfig:"OS_IMAGE_DOWNLOAD_TRUSTED_CA_FILE" default:""`
+
+	// This is a path to a CA file that will be trusted for TLS connections to the Assisted Service API
+	// this will be used for API calls back to the Assisted Service API
+	// Will default to the value held in HTTPS_CA_FILE unless overridden
+	AssistedServiceApiTrustedCAFile string `envconfig:"ASSISTED_SERVICE_API_TRUSTED_CA_FILE" default:""`
 }
 
 func main() {
+	if Options.HTTPSCAFile != "" {
+		Options.AssistedServiceApiTrustedCAFile = Options.HTTPSCAFile
+	}
 	log.SetReportCaller(true)
 	log.SetFormatter(&log.JSONFormatter{})
 	err := envconfig.Process("cluster-image", &Options)
@@ -67,7 +82,7 @@ func main() {
 		}
 	}
 
-	is, err := imagestore.NewImageStore(isoeditor.NewEditor(Options.DataDir), Options.DataDir, Options.ImageServiceBaseURL, Options.InsecureSkipVerify, versions)
+	is, err := imagestore.NewImageStore(isoeditor.NewEditor(Options.DataDir), Options.DataDir, Options.ImageServiceBaseURL, Options.InsecureSkipVerify, versions, Options.AssistedServiceApiTrustedCAFile)
 	if err != nil {
 		log.Fatalf("Failed to create image store: %v\n", err)
 	}
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -3,6 +3,7 @@ package imagestore
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"net/http"
@@ -84,7 +85,7 @@ const (
 	ImageTypeMinimal = "minimal-iso"
 )
 
-func NewImageStore(ed isoeditor.Editor, dataDir, imageServiceBaseURL string, insecureSkipVerify bool, versions []map[string]string) (ImageStore, error) {
+func NewImageStore(ed isoeditor.Editor, dataDir, imageServiceBaseURL string, insecureSkipVerify bool, versions []map[string]string, osImageDownloadTrustedCAFile string) (ImageStore, error) {
 	if err := validateVersions(versions); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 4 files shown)


gamli75:
@omertuc @carbonin please review

omertuc:
PEM certificates can be bundled into a single file, typically programs accept just a single file, not an entire directory of files,  I think it's better if we follow this convention as well, as it also makes things simpler for us

pkg/imagestore/imagestore.go
```diff
@@ -84,16 +86,65 @@ const (
 	ImageTypeMinimal = "minimal-iso"
 )
 
-func NewImageStore(ed isoeditor.Editor, dataDir, imageServiceBaseURL string, insecureSkipVerify bool, versions []map[string]string) (ImageStore, error) {
+func getAdditionalTLSCertPaths(additionalCAPath string) ([]string, error) {
+	if strings.Trim(additionalCAPath, " ") == "" {
```
omertuc, line 90: ` ` is a valid file name, no need to do trimming here. If someone is putting spaces in their environment variables, I'd like to tell them about it with an explicit error of "file not found", not pretend that everything is fine

Dockerfile.image-service
```diff
@@ -13,5 +13,8 @@ RUN mkdir $DATA_DIR && chmod 775 $DATA_DIR
 VOLUME $DATA_DIR
 ENV DATA_DIR=$DATA_DIR
 
+RUN mkdir /additional-ca-bundles && chmod 775 /additional-ca-bundles
+VOLUME /additional-ca-bundles
```
omertuc, line 17: What purpose does this serve?

paul-maidment:
> PEM certificates can be (and often are) bundled into a single file, typically programs accept just a single file, not an entire directory of files, I think it's better if we follow this convention as well, as it also makes things simpler for us

I am trying to maintain consistency with the existing approach used by the Openshift CA controller. 

The instruction given to the customer for the creation of a ConfigMap containing a single file (manually) would be something like 

`oc -n assisted-installer create configmap image-service-additional-ca --from-file=ca-out/ca.crt`

There is however, nothing to stop the user from doing...

`oc -n assisted-installer create configmap image-service-additional-ca --from-file=ca-out/ca.crt --from-file=ca-out/some-other-ca.crt`

The existing implementation works well and is flexible enough to allow PEM certificates to be added either as a single concatenated file or as multiple (possibly also concatenated files)

I argue that this is not a problem for the following reasons 

- The customer may not wish to concatenate all of their PEMs together for whatever reason, it may be simpler for them to use the multi file approach above.
- I can't see a significant disadvantage to this approach, only benefits and flexibility
- It's good enough for the Openshift CA controller to take this approch

main.go
```diff
@@ -37,6 +37,7 @@ var Options struct {
 	InsecureSkipVerify    bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
 	ImageServiceBaseURL   string `envconfig:"IMAGE_SERVICE_BASE_URL"`
 	LogLevel              string `envconfig:"LOGLEVEL" default:"info"`
+	AdditionalCAFile      string `envconfig:"ADDITIONAL_CA_FILE" default:""`
```
omertuc, line 40: This name is not great, we already have `HTTPS_CA_FILE` a few lines above. Please choose a different name and document the difference between them

pkg/imagestore/imagestore.go
```diff
@@ -84,16 +85,34 @@ const (
...
 	}
 	transportConfig, ok := http.DefaultTransport.(*http.Transport)
 	if !ok {
 		return nil, fmt.Errorf("expected http.DefaultTransport to be of type *http.Transport")
 	}
+
 	myTransport := transportConfig.Clone()
 	myTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: insecureSkipVerify} //nolint:gosec // Optionally ignore TLS (G402 error)
+
+	// Add additional TLS certificates (if available) for pulling images.
```
omertuc, line 100: The verb `pulling` implies that you're talking about a container registry. This is not accessing a container registry. This is accessing an HTTP server serving OS ISO images over raw HTTP

pkg/imagestore/imagestore.go
```diff
@@ -84,16 +85,34 @@ const (
...
+	if additionalCAFile != "" {
+		caCertPool := x509.NewCertPool()
+		additionalTLSCert, err := os.ReadFile(additionalCAFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open additional certificate file %s, %w", additionalCAFile, err)
+		}
+		if !caCertPool.AppendCertsFromPEM(additionalTLSCert) {
+			return nil, fmt.Errorf("failed to append additional certificate to pool %s", additionalCAFile)
```
omertuc, line 108:
```suggestion
			return nil, fmt.Errorf("failed to append additional certificate %s to pool", additionalCAFile)
```

main.go
```diff
@@ -22,21 +22,22 @@ import (
...
-	AllowedDomains        string `envconfig:"ALLOWED_DOMAINS"`
-	InsecureSkipVerify    bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
-	ImageServiceBaseURL   string `envconfig:"IMAGE_SERVICE_BASE_URL"`
-	LogLevel              string `envconfig:"LOGLEVEL" default:"info"`
+	AssistedServiceScheme        string `envconfig:"ASSISTED_SERVICE_SCHEME"`
+	AssistedServiceHost          string `envconfig:"ASSISTED_SERVICE_HOST"`
+	DataDir                      string `envconfig:"DATA_DIR"`
+	HTTPSKeyFile                 string `envconfig:"HTTPS_KEY_FILE"`
+	HTTPSCertFile                string `envconfig:"HTTPS_CERT_FILE"`
+	HTTPSCAFile                  string `envconfig:"HTTPS_CA_FILE"`
```
omertuc, line 30: Let's change this name too:

- Create a new param with a new env var name
- Copy the value from HTTPS_CA_FILE into the new var for backwards compatibility

main.go
```diff
@@ -22,24 +22,41 @@ import (
...
-	LogLevel              string `envconfig:"LOGLEVEL" default:"info"`
+	AssistedServiceScheme           string `envconfig:"ASSISTED_SERVICE_SCHEME"`
+	AssistedServiceHost             string `envconfig:"ASSISTED_SERVICE_HOST"`
+	DataDir                         string `envconfig:"DATA_DIR"`
+	HTTPSKeyFile                    string `envconfig:"HTTPS_KEY_FILE"`
+	HTTPSCertFile                   string `envconfig:"HTTPS_CERT_FILE"`
+
+	// Legacy/Depracted - Whenever possible please use ASSISTED_SERVICE_API_TRUSTED_CA_FILE
+	// This is a path to a CA file that can be used for TLS connections to the Assisted Service API
+	// this will be used for API calls back to the Assisted Service API
```
omertuc, line 33:
```suggestion
	// Depracted - use ASSISTED_SERVICE_API_TRUSTED_CA_FILE instead
```

main.go
```diff
@@ -22,24 +22,41 @@ import (
...
+	HTTPListenPort                  string `envconfig:"HTTP_LISTEN_PORT"`
+	MaxConcurrentRequests           int64  `envconfig:"MAX_CONCURRENT_REQUESTS" default:"400"`
+	RHCOSVersions                   string `envconfig:"RHCOS_VERSIONS"`
+	OSImages                        string `envconfig:"OS_IMAGES"`
+	AllowedDomains                  string `envconfig:"ALLOWED_DOMAINS"`
+	InsecureSkipVerify              bool   `envconfig:"INSECURE_SKIP_VERIFY" default:"false"`
+	ImageServiceBaseURL             string `envconfig:"IMAGE_SERVICE_BASE_URL"`
+	LogLevel                        string `envconfig:"LOGLEVEL" default:"info"`
+
+	// This is a path to a CA file that will be used for fetching OS Images
```
omertuc, line 46:
```suggestion
	// This is a path to a CA file that will be trusted when fetching OS Images
```

main.go
```diff
@@ -22,24 +22,41 @@ import (
...
+	// intended for scenarios where the customer is using an external service to host OS images.
```
omertuc, line 47: `user` preferred over `customer`

```suggestion
	// intended for scenarios where the OS images are served from a service that uses a custom CA
```

main.go
```diff
@@ -22,24 +22,41 @@ import (
...
+	OSImageDownloadTrustedCAFile    string `envconfig:"OS_IMAGE_DOWNLOAD_TRUSTED_CA_FILE" default:""`
+
+	// This is a path to a CA file that can be used for TLS connections to the Assisted Service API
```
omertuc, line 50:
```suggestion
	// This is a path to a CA file that will be used for TLS connections to the Assisted Service API
```

omertuc:
Will `/lgtm` once the commit message is updated to reflect all the changes

PR #185, opened 2024-03-15
MULTIARCH-2678, OCPBUGS-14478: Add a NewIgnitionImageReader() function

zaneb:
Add a `NewIgnitionImageReader()` function to the `isoeditor` package that will allow the agent installer to generate the correct data for an ignition image:

- without knowing the filename (which is different on different architectures)
- even if there is other data apart from the embed area in the image (which happens on s390)
- even if the ignition data is larger than the embed area (as long as it is safe to do so)
- passing the type-safe IgnitionContent type
- reusing the same code that `NewRHCOSStreamReader()` uses to find the embed area

The function returns the filename within the ISO and a Reader for the data. This can be copied over the corresponding file in the directory unpacked by `Extract()`.

### DIFF
--- a/pkg/isoeditor/ignition_image.go
+++ b/pkg/isoeditor/ignition_image.go
@@ -1,0 +1,41 @@
+package isoeditor
+
+import "io"
+
+type FileData struct {
+	Filename string
+	Data     io.ReadCloser
+}
+
+// NewIgnitionImageReader returns the filename of the ignition image in the ISO,
+// along with a stream of the ignition image with ignition content embedded.
+// This can be used to overwrite the ignition image file of an ISO previously
+// unpacked by Extract() in order to embed ignition data.
+func NewIgnitionImageReader(isoPath string, ignitionContent *IgnitionContent) (FileData, error) {
+	info, iso, err := ignitionOverlay(isoPath, ignitionContent)
+	if err != nil {
+		return FileData{}, err
+	}
+	imageOffset, imageLength, err := GetISOFileInfo(info.File, isoPath)
+	if err != nil {
+		return FileData{}, err
+	}
+
+	if _, err := iso.Seek(imageOffset, io.SeekStart); err != nil {
+		iso.Close()
+		return FileData{}, err
+	}
+	data := struct {
+		io.Reader
+		io.Closer
+	}{
+		Reader: io.LimitReader(iso, imageLength),
+		Closer: iso,
+	}
+
+	ignitionImage := FileData{
+		Filename: info.File,
+		Data:     data,
+	}
+	return ignitionImage, nil
+}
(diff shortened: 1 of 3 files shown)


pkg/isoeditor/ignition_image.go
```diff
@@ -0,0 +1,39 @@
...
+	info, iso, err := ignitionOverlay(isoPath, ignitionContent, true)
+	if err != nil {
+		return "", nil, err
+	}
+	imageOffset, imageLength, err := GetISOFileInfo(info.File, isoPath)
+	if err != nil {
+		return "", nil, err
+	}
+
+	length := info.Offset + info.Length
```
carbonin, line 24: The length for `LimitReader` should be the number of bytes you want to read, not an absolute location. Why include the offset here?

zaneb: The total length of the image file is the longer of:

- the length of the header before the embed area within the image (info.Offset) plus the length of the embed area (info.Length); or
- the length of the original image file

The former will be longer if we have expanded the image to fit more ignition data.
The latter will be longer if there is a trailer after the embed area.
Mostly they'll be the same.

carbonin: To amend my previous comment, but also to respond to yours, I'm pretty sure `info.Offset` is the offset within the named file in the ignition info where the embed area starts.

So this function will serve the entire file within which the ignition is embedded whether the ignition archive is the whole file or not?

IIRC this internal file reference was created for s390x arch. What even is the file it's embedded in in that case and why do you want access to it in this way?

carbonin: I guess so you can just blast the entire extracted file rather than writing into it at the offset?

zaneb: > So this function will serve the entire file within which the ignition is embedded whether the ignition archive is the whole file or not?

Yep, exactly. It returns the filename and the new content of the file, so we can overwrite that file. Everything is abstracted except `os.Open()` and `ReadFrom()`, so there's almost no way for the caller to mess it up.

pkg/isoeditor/ignition_image.go
```diff
@@ -0,0 +1,68 @@
...
+type ignitionImageReader struct {
+	io.Reader
+	io.Closer
+}
+
+// NewIgnitionImageReader returns the filename of the ignition image in the ISO,
+// along with a stream of the ignition image with ignition content embedded.
+// This can be used to overwrite the ignition image file of an ISO previously
+// unpacked by Extract() in order to embed ignition data.
+func NewIgnitionImageReader(isoPath string, ignitionContent *IgnitionContent) ([]FileData, error) {
```
carbonin, line 23: Looks good, but we should probably add a test of some sort for this.

zaneb: Done

carbonin (approved):

PR #190, opened 2024-03-25
OCPBUGS-14478: Pad expanded ignition embed area

zaneb:
If the ignition embed area is expanded in the ignition image returned by `NewIgnitionImageReader()`, also add additional padding at the end of the image. This allows the ignition image to be extracted by coreos-installer and re-inserted into the image after minor changes that may make it slightly larger.

Currently the embed area is 256kB. This change ensures that the size of the image returned is a multiple of 32kB, and that there is also at least 32kB free after the provided ignition. This size is proportional to the embed area (i.e. we always leave 1/4 of the original embed area free).

### DIFF
--- a/pkg/isoeditor/stream.go
+++ b/pkg/isoeditor/stream.go
@@ -131,8 +131,11 @@ func (ibf *ignitionBoundaryFinder) findBoundaries(filePath, isoPath string) (int
 	// allow overflow if requested and if the embed area extends all the way
 	// to the end of the file
 	if ibf.allowOverflow && ((info.Offset + info.Length) >= isoFileLength) {
-		if ibf.dataSize > info.Length {
-			info.Length = ibf.dataSize
+		chunkSize := (info.Length + 3) / 4
+		if (ibf.dataSize + chunkSize) > info.Length {
+			// increase size in chunks equal to a quarter of the original size,
+			// ensuring that there is always at least one full chunk free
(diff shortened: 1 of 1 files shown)


pkg/isoeditor/stream.go
```diff
@@ -75,6 +75,26 @@ func ignitionOverlay(isoPath string, ignitionContent *IgnitionContent, allowOver
...
+	if ibf.info.Length > ibf.dataSize {
+		offset, _, err := GetISOFileInfo(ibf.info.File, isoPath)
+		if err != nil {
+			r.Close()
+			return nil, nil, err
+		}
+		paddingLen := ibf.info.Length - ibf.dataSize
+		paddingOverlay := overlay.Overlay{
+			Reader: bytes.NewReader(bytes.Repeat([]byte{0}, int(paddingLen))),
+			Offset: offset + ibf.info.Offset + ibf.dataSize,
```
carbonin, line 88: I may be misunderstanding, but this isn't safe to do in a streaming context, right?
Wouldn't we need to expand the size of the entire ISO to ensure that nothing was overwritten after the ignition embed area?

Maybe this will only ever be done if we've already set `allowOverflow`, but that's not clear from this new code.

zaneb: It's always safe to overwrite from the end of the data to the end of the embed area with zeroes. That's what this commit does.

It's only safe to to extend the embed area if allowOverflow is true, but that is handled by the other commit in findBoundaries().

carbonin (approved):

PR #191, opened 2024-03-25
AGENT-890: Add NewKargsReader() function

zaneb:
Add a function analagous to the `NewIgnitionImageReader()` function, but for kargs. Given a path to an ISO and some kargs to append, it returns a list of files and the new contents that they need to be overwritten with. This will allow the agent-based installer to stop duplicating the code to append kargs in the ISO.

### DIFF
--- a/pkg/isoeditor/isolate_file.go
+++ b/pkg/isoeditor/isolate_file.go
@@ -1,0 +1,38 @@
+package isoeditor
+
+import (
+	"io"
+
+	"github.com/openshift/assisted-image-service/pkg/overlay"
+)
+
+type FileData struct {
+	Filename string
+	Data     io.ReadCloser
+}
+
+func isolateISOFile(isoPath, file string, data overlay.OverlayReader, minLength int64) (FileData, bool, error) {
+	fileOffset, fileLength, err := GetISOFileInfo(file, isoPath)
+	if err != nil {
+		return FileData{}, false, err
+	}
+
+	expanded := false
+	if minLength > fileLength {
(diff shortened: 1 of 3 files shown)


pkg/isoeditor/isolate_file.go
```diff
@@ -0,0 +1,38 @@
...
+	fileOffset, fileLength, err := GetISOFileInfo(file, isoPath)
+	if err != nil {
+		return FileData{}, false, err
+	}
+
+	expanded := false
+	if minLength > fileLength {
+		fileLength = minLength
+		expanded = true
+	}
```
carbonin, line 24: This is not what this condition was doing previously IIUC.

Before this was checking if the file withing which the ignition existed had any additional data outside of the ignition, now it's checking if the data we want to embed is larger than the file we need to embed it into.

Did you intentionally change this?

zaneb: They should be equivalent.

We want the length returned to be the larger of (a) the original file length, or (b) the new file length after expansion (= the offset of the embed area + the new length of the embed area). (The former could be larger if there is a trailer after the embed area. Obviously if they're both the same size then it doesn't matter which we choose.)

For convenience, the condition here has flipped from effectively `max(info.Offset + info.Length, imageLength)` to `max(fileLength, info.Offset + info.Length)` but there should be no change in behaviour.

carbonin: Ah okay, I see now, thanks for the explanation.

carbonin (approved):

PR #255, opened 2024-07-06
[release-ocm-2.8] MGMT-18332: Download rpm packages in a different stage at the build of Dockerfile-build

danmanor:
Download rpm packages in a different stage at the build of Dockerfile-build as golang image is based on centos 7 which cannot reach its repositories as it is EOL

### DIFF
--- a/Dockerfile.image-service-build
+++ b/Dockerfile.image-service-build
@@ -1,6 +1,27 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.18
-ENV GO111MODULE=on
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS golang
+
 ENV GOFLAGS=""
 
-COPY hack/setup_env.sh ./
-RUN ./setup_env.sh
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.46.0 && \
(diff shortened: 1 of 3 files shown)


Dockerfile.image-service-build
```diff
@@ -1,6 +1,27 @@
...
-ENV GO111MODULE=on
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 AS golang
+
 ENV GOFLAGS=""
 
-COPY hack/setup_env.sh ./
-RUN ./setup_env.sh
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.46.0
+RUN go install golang.org/x/tools/cmd/goimports@v0.1.5
+RUN go install github.com/golang/mock/mockgen@v1.6.0
```
rccrdpccl, line 7: these can run as a single RUN to reduce number of layers

Dockerfile.image-service-build
```diff
@@ -1,6 +1,27 @@
...
+FROM quay.io/centos/centos:stream9
+
+RUN dnf install -y \
+        podman \ 
+        genisoimage \ 
+        make \
+        git \ 
+        gcc \
+        openssl-devel
+RUN dnf clean all
```
rccrdpccl, line 18: Same here: we can reduce number of RUN statements to run as a single layer

PR #253, opened 2024-07-05
[release-ocm-2.9] NO-ISSUE: Align build dockerfile with centos9

pastequo:
## Description

Align build dockerfile to use go-toolset for go binary and centos9 for dependencies

setup_env.sh had been removed from prow release so I think it could be removed

## How was this code tested?

running various skipper command

## Assignees

/cc @danmanor 
/cc @

## Checklist

- [x] Title and description added to both, commit and PR
- [x] Relevant issues have been associated
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit tests (note that code changes require unit tests)

PR #256, opened 2024-07-06
[release-ocm-2.9] MGMT-18332: Download rpm packages in a different stage at the build of Dockerfile-build

danmanor:
Download rpm packages in a different stage at the build of Dockerfile-build as golang image is based on centos 7 which cannot reach its repositories as it is EOL


pastequo:
Compared to this one https://github.com/openshift/assisted-image-service/pull/253, here we are keeping the hack file (and the associated Makefile target which isn't called anymore in prow pipeline).

Any opinion on that ?

danmanor:
thanks @pastequo

PR #266, opened 2024-07-24
MGMT-18313: Replace golang base image as it is based on Centos Linux 7

danmanor:


adriengentil (approved):
We will need to cleanup Prow config as well: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/assisted-image-service/openshift-assisted-image-service-main__edge.yaml#L57-L62

danmanor:
@adriengentil  Yes

PR #274, opened 2024-08-22
MGMT-18578 - Add nmstatectl binary to initrd in minimal ISO for OCP 4.13-4.17 (x86_64 architecture)

linoyaslan:
## Description

As part of integrating the nmstate service into our flow, it's necessary to include the `nmstatectl` binary in the initrd for minimal ISO cases. This ensures that the pre-network-manager-config service can run the nmstate service early during the boot process, allowing the nmstate YAML to be applied as keyfiles.
Note: In a later phase, I will fetch the binary from the rootfs. This PR is just the initial implementation. See - [MGMT-18577](https://issues.redhat.com/browse/MGMT-18577)

## How was this code tested?

To verify this, I used test-infra's Minikube, generated and downloaded the minimal ISO, then interrupted the boot process to confirm that the `nmstate.img` was successfully added to the initrd.

## Assignees

/cc @danielerez 
/cc @carbonin 
/cc @jhernand 
/cc @AlonaKaplan 

## Checklist

### DIFF
--- a/pkg/isoeditor/rhcos.go
+++ b/pkg/isoeditor/rhcos.go
@@ -5,13 +5,16 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 
 	log "github.com/sirupsen/logrus"
 )
 
 const (
 	RamDiskPaddingLength = uint64(1024 * 1024) // 1MB
+	NmstatectlPath       = "/usr/bin/nmstatectl"
 	ramDiskImagePath     = "/images/assisted_installer_custom.img"
+	nmstateDiskImagePath = "/images/nmstate.img"
 )
 
 //go:generate mockgen -package=isoeditor -destination=mock_editor.go . Editor
@@ -20,11 +23,14 @@ type Editor interface {
 }
 
 type rhcosEditor struct {
-	workDir string
+	workDir, nmstatectlPath string
 }
 
-func NewEditor(dataDir string) Editor {
-	return &rhcosEditor{workDir: dataDir}
+func NewEditor(dataDir, nmstatectlPath string) Editor {
+	return &rhcosEditor{
+		workDir:        dataDir,
+		nmstatectlPath: nmstatectlPath,
+	}
 }
 
 // CreateMinimalISO Creates the minimal iso by removing the rootfs and adding the url
@@ -38,14 +44,19 @@ func CreateMinimalISO(extractDir, volumeID, rootFSURL, arch, minimalISOPath stri
 		return err
 	}
 
-	if err := fixGrubConfig(rootFSURL, extractDir); err != nil {
+	var includeNmstateRamDisk bool
+	if _, err := os.Stat(filepath.Join(extractDir, nmstateDiskImagePath)); err == nil {
+		includeNmstateRamDisk = true
+	}
+
+	if err := fixGrubConfig(rootFSURL, extractDir, includeNmstateRamDisk); err != nil {
 		log.WithError(err).Warnf("Failed to edit grub config")
 		return err
 	}
 
 	// ignore isolinux.cfg for ppc64le because it doesn't exist
 	if arch != "ppc64le" {
-		if err := fixIsolinuxConfig(rootFSURL, extractDir); err != nil {
+		if err := fixIsolinuxConfig(rootFSURL, extractDir, includeNmstateRamDisk); err != nil {
 			log.WithError(err).Warnf("Failed to edit isolinux config")
 			return err
 		}
@@ -73,7 +84,44 @@ func (e *rhcosEditor) CreateMinimalISOTemplate(fullISOPath, rootFSURL, arch, min
 		return err
 	}
 
+	ramDiskPath := filepath.Join(extractDir, nmstateDiskImagePath)
+
+	// Ensure that the runtime CPU arch equals to the requested arch
+	if normalizeCPUArchitecture(runtime.GOARCH) == arch {
+		err = createNmstateRamDisk(arch, e.nmstatectlPath, ramDiskPath)
+		if err != nil {
+			return fmt.Errorf("failed to create nmstate ram disk for arch %s: %v", arch, err)
+		}
+	}
+
 	err = CreateMinimalISO(extractDir, volumeID, rootFSURL, arch, minimalISOPath)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createNmstateRamDisk(arch, nmstatectlPath, ramDiskPath string) error {
+	// Check if nmstatectl binary file exists
+	if _, err := os.Stat(nmstatectlPath); os.IsNotExist(err) {
+		return err
+	}
+
+	// Read binary
+	nmstateBinContent, err := os.ReadFile(nmstatectlPath)
+	if err != nil {
+		return err
+	}
+
+	// Create a compressed RAM disk image with the nmstatectl binary
+	compressedCpio, err := generateCompressedCPIO(nmstateBinContent, nmstatectlPath, 0o100_755)
+	if err != nil {
+		return err
+	}
+
+	// Write RAM disk file
+	err = os.WriteFile(ramDiskPath, compressedCpio, 0755) //nolint:gosec
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


jhernand:
Putting the binary inside the repository and inside the image of the service may be problematic: it will be very hard to replace it if we find a bug, for example. I know that eventually it will be taken from the rootfs, but could we meanwhile take it from some configurable place outside of the image? For example, could we have a new `EXTRA_INITRD_FILES` environment pointing to a directory containing them? Maybe with sub-directories for different architectures? I mean something like this:

```
$ find "${EXTRA_INITRD_FILES}" -type f
.../x86_64/usr/bin/nmstatectl
.../arm64/usr/bin/nmstatectl
```

Then providing the files will be done when the service is deployed, not when it is built.

I'd also suggest to make this optional: if the variable isn't set then the behavior should be to not include the new initrd in the ISO. That will help with upgrades of existing environments.

Finally I'd suggest to give a more generic name to the initrd file, something like `extra.img`, because that way we can reuse it for other files not related to nmstate.

danielerez:
> Putting the binary inside the repository and inside the image of the service may be problematic: it will be very hard to replace it if we find a bug, for example. I know that eventually it will be taken from the rootfs, but could we meanwhile take it from some configurable place outside of the image? For example, could we have a new `EXTRA_INITRD_FILES` environment pointing to a directory containing them? Maybe with sub-directories for different architectures? I mean something like this:
> 
> ```
> $ find "${EXTRA_INITRD_FILES}" -type f
> .../x86_64/usr/bin/nmstatectl
> .../arm64/usr/bin/nmstatectl
> ```
> 
> Then providing the files will be done when the service is deployed, not when it is built.
> 
> I'd also suggest to make this optional: if the variable isn't set then the behavior should be to not include the new initrd in the ISO. That will help with upgrades of existing environments.
> 

Hey @jhernand, so with this approach it means we'd need to manually add the `nmstatectl` binary on SaaS, right? Which means one more thing to maintain :/
Maybe, as an alternative, for now we can simply amend the code to download the binary: https://github.com/nmstate/nmstate/releases/download/v2.2.35/nmstatectl-linux-x64.zip
It could be of course cached, as we do for the ISOs.
wdyt?

carbonin:
>Maybe, as an alternative, for now we can simply amend the code to download the binary: https://github.com/nmstate/nmstate/releases/download/v2.2.35/nmstatectl-linux-x64.zip

How does this work for disconnected on-prem installs?

carbonin:
If we're only doing x86 could we install the RPM in the container build and use that?
That would also allow us to take advantage of security scanning for images. Also on the security note I think it's generally not good practice to include non-source-based binaries in images. I'm not sure if anything in our build pipeline will block us though as I can't remember where I read this.

carbonin:
If not the RPM approach (and we don't plan the committed-binary to prod) is there some reason to not be doing the download in the container build where we would at least be able to see how the binary is downloaded and what version it is?

jhernand:
My understanding is that as @linoyaslan says in the description this is a temporary implementation to ease further development, and in the future (near future, I hope) we will change this code to extract the binary from the `rootfs.img` file. What I suggest is to let this pull request move forward, with the capability to enable/disable this functionality at will, and without including the binaries in the repository or the image.

Downloading the file in runtime like @danielerez suggests seems reasonable to me as well, as long we understand that is also a temporary implementation that can be enabled/disabled.

Installing the RPM, even if it is in the container build, also makes it hard to replace those files without upgrading the service, but it is better than having the binary in the repository.

Or maybe we should just wait till the `rootfs.img` implementation is ready?

pkg/isoeditor/rhcos.go
```diff
@@ -27,12 +32,77 @@ func NewEditor(dataDir string) Editor {
...
+
+	// Check if the file exists
+	if _, err := os.Stat(srcPath); os.IsNotExist(err) {
+		log.Fatalf("nmstatectl_x86_64_v4.17 source file does not exist: %v", err)
+	}
+
+	nmstateBinContent, err := os.ReadFile(srcPath)
+	if err != nil {
+		fmt.Println("Error reading file:", err)
+		return []byte{}, err
```
carbonin, line 46: You can use `nil` for the slice return value in an error case.

pkg/isoeditor/rhcos.go
```diff
@@ -27,12 +32,77 @@ func NewEditor(dataDir string) Editor {
...
+	if err := gzipWriter.Close(); err != nil {
+		return nil, errors.Wrap(err, "Failed to close gzip writer")
+	}
+
+	padSize := (4 - (compressedBuffer.Len() % 4)) % 4
+	for i := 0; i < padSize; i++ {
+		if err := compressedBuffer.WriteByte(0); err != nil {
+			return nil, err
+		}
+	}
```
carbonin, line 79: Can we create a helper for this? It's not something that I feel like should be repeated.

Maybe even create some way to dedupe even more of this function and [`Archive`](https://github.com/openshift/assisted-image-service/blob/180a34e65986e17b3ee16861ca40d64ae442f0d6/pkg/isoeditor/ignition.go#L15-L48)?

carbonin (changes requested):
We should also add the additional initrd in the isolinux case.

linoyaslan:
@jhernand @carbonin  I would appreciate your review again.

pkg/isoeditor/rhcos.go
```diff
@@ -129,7 +169,7 @@ func fixGrubConfig(rootFSURL, extractDir string) error {
 	}
 
 	// Edit config to add custom ramdisk image to initrd
-	if err := editFile(foundGrubPath, `(?m)^(\s+initrd) (.+| )+$`, fmt.Sprintf("$1 $2 %s", ramDiskImagePath)); err != nil {
+	if err := editFile(foundGrubPath, `(?m)^(\s+initrd) (.+| )+$`, fmt.Sprintf("$1 $2 %s %s", ramDiskImagePath, nmstateDiskImagePath)); err != nil {
```
danielerez, line 172: I think it should be added only if the nmstate disk image was generated. Boot would probably fail with a missing ram disk file.

pkg/isoeditor/rhcos.go
```diff
@@ -146,7 +186,7 @@ func fixIsolinuxConfig(rootFSURL, extractDir string) error {
 		return err
 	}
 
-	if err := editFile(filepath.Join(extractDir, "isolinux/isolinux.cfg"), `(?m)^(\s+append.*initrd=\S+) (.*)$`, fmt.Sprintf("${1},%s ${2}", ramDiskImagePath)); err != nil {
+	if err := editFile(filepath.Join(extractDir, "isolinux/isolinux.cfg"), `(?m)^(\s+append.*initrd=\S+) (.*)$`, fmt.Sprintf("${1},%s,%s ${2}", ramDiskImagePath, nmstateDiskImagePath)); err != nil {
```
danielerez, line 189: Same, should be added only if `nmstateDiskImagePath` exists.

pkg/isoeditor/rhcos_test.go
```diff
@@ -55,52 +56,152 @@ var _ = Context("with test files", func() {
...
-		validateFileContainsLine(filepath.Join(filesDir, "EFI/redhat/grub.cfg"), grubCfg)
+			ramDisk, err := os.CreateTemp(extractDir, "nmstate.img")
+			Expect(err).ToNot(HaveOccurred())
+			ramDiskPath = ramDisk.Name()
 
+			arch = normalizeCPUArchitecture(runtime.GOARCH)
+		})
+
+		AfterEach(func() {
+			os.Remove(nmstatectlPath)
```
carbonin, line 129: You should be removing the entire `extractDir` path here. Otherwise I think you'll end up leaking temp dirs.

pkg/isoeditor/rhcos.go
```diff
@@ -81,6 +98,38 @@ func (e *rhcosEditor) CreateMinimalISOTemplate(fullISOPath, rootFSURL, arch, min
 	return nil
 }
 
+func createNmstateRamDisk(arch, nmstatectlPath, ramDiskPath string) error {
+	// Ensure that the runtime CPU arch equals to the requested arch
+	if normalizeCPUArchitecture(runtime.GOARCH) != arch {
+		return nil
```
carbonin, line 104: I think this is a bit unexpected.

I'd rather `createNmstateRamDisk` fail if it doesn't ... create the nmstate ram disk :shrug: 

I'd say do this arch check in the caller and either don't call this function at all or error when the architecture doesn't match.

carbonin (commented):
Few minor comments, otherwise looks good.

PR #280, opened 2024-10-06
MGMT-18577: fetch nmstate binary from rootfs

linoyaslan:
## Description
Following [this PR](https://github.com/openshift/assisted-image-service/pull/274) , the next step is to fetch the nmstate binary from the rootfs instead of using the nmstate binary we installed in the image-service pod.

## Assignees

/cc @danielerez 
/cc @carbonin @jhernand 

## Checklist


linoyaslan:
@carbonin @jhernand  I’d appreciate your feedback on this solution. I’ve tried other approaches, like using the `go-diskfs` package, but had issues with squashfs. Once we agree on this approach, I’ll proceed with adding the appropriate unit tests.

Thank you!

pkg/isoeditor/rhcos.go
```diff
@@ -68,8 +74,60 @@ func CreateMinimalISO(extractDir, volumeID, rootFSURL, arch, minimalISOPath stri
...
+		return "", err
+	}
+	rootfsPath := filepath.Join(extractDir, "images/pxeboot/rootfs.img")
+	_, err = execute(fmt.Sprintf("7z x %s", rootfsPath), nmstateDir)
+	if err != nil {
+		log.Errorf("failed to 7z x rootfs.img: %v", err.Error())
+		return "", err
+	}
+	// limiting files is needed on el<=9 due to https://github.com/plougher/squashfs-tools/issues/125
+	ulimit := "ulimit -n 1024"
```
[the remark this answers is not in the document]
jhernand: In case it helps: the `ulimit -n ...` builtin command only changes the limit for process of the shell that runs it, and for the processes created by that shell. As the code below only uses it before running `unsquashfs ...` it will have no effect in any other thing that uses this image.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+//go:generate mockgen -package=isoeditor -destination=mock_nmstate_handler.go . NmstateHandler
+type NmstateHandler interface {
+	CreateNmstateRamDisk(nmstatectlPath, ramDiskPath string) error
+	ExtractNmstatectl(rootfsPath string) (string, error)
```
carbonin, line 19: The way these functions are written looks like they shouldn't be separate.

Specifically how one creates a tempdir and the other cleans it up is liable to leak resources when things go wrong.

If you want to split them out for clarity in the private implementation, that's fine, but ensure you're managing the `nmstateDir` correctly.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,114 @@
...
+	if err := os.RemoveAll(n.nmstateDir); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (n *nmstateHandler) ExtractNmstatectl(rootfsPath string) (string, error) {
+	var err error
+	n.nmstateDir, err = os.MkdirTemp(n.workDir, "nmstate")
```
carbonin, line 66: This doesn't feel like it should be state.

This makes the object not thread safe and also looks like this will leak this directory on error in this function or `CreateNmstateRamDisk`

carbonin (changes requested):
Can we get an idea of how much (if any) additional space this will require when the image service is starting up?

It looks like we're not creating the minimal ISOs in parallel any more, so it might not but as much of an issue, but it would still be good to know.

linoyaslan:
> Can we get an idea of how much (if any) additional space this will require when the image service is starting up?
> 
> It looks like we're not creating the minimal ISOs in parallel any more, so it might not but as much of an issue, but it would still be good to know.

@carbonin Extracting the rootfs is approximately 1 GB.

carbonin:
Also I wonder if startup time is affected.

@rccrdpccl what do you think?

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (n *nmstateHandler) extractNmstatectl(rootfsPath, nmstateDir string) (string, error) {
+	_, err := n.executer.Execute(fmt.Sprintf("7z x %s", rootfsPath), nmstateDir)
+	if err != nil {
+		log.Errorf("failed to 7z x rootfs.img: %v", err.Error())
+		return "", err
+	}
+	// limiting files is needed on el<=9 due to https://github.com/plougher/squashfs-tools/issues/125
+	ulimit := "ulimit -n 1024"
+	list, err := n.executer.Execute(fmt.Sprintf("%s ; unsquashfs -d '' -lc %s", ulimit, "root.squashfs"), nmstateDir)
```
carbonin, line 86: What is `-lc`? I don't see a `-lc` or `-c` option in `man unsquashfs` on my system.

Also does  `-d ''` unpack into the working directory?
Maybe just add a comment explaining the `unsquashfs` options you're using here.

carbonin: Also maybe use the non-abbreviated version on the options like you do on the later command.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,127 @@
...
+		return err
+	}
+
+	// Write RAM disk file
+	err = os.WriteFile(ramDiskPath, compressedCpio, 0755) //nolint:gosec
+	if err != nil {
+		return err
+	}
+
+	// Remove temp dir
```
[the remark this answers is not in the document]
danielerez: +1, makes sense.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,134 @@
...
+		log.Errorf("failed to 7z x rootfs.img: %v", err.Error())
+		return "", err
+	}
+	// limiting files is needed on el<=9 due to https://github.com/plougher/squashfs-tools/issues/125
+	ulimit := "ulimit -n 1024"
+
+	// Listing the filesystem concisely, displaying only files (using `-lc` option).
+	// Each file in the output won't include any prefix before `/ostree` (by using `-dest ''` option),
+	// which is useful when invoking `-extract-file` (after finding the `nmstatectl` binary path).
+	list, err := n.executer.Execute(fmt.Sprintf("%s ; unsquashfs -dest '' -lc %s", ulimit, "root.squashfs"), nmstateDir)
```
rccrdpccl, line 93: Any reason why we picked `unsquashfs` over go pkg `go-diskfs`'s squashfs?

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,134 @@
...
+	_, err = n.executer.Execute(fmt.Sprintf("%s ; unsquashfs -no-xattrs %s -extract-file %s", ulimit, "root.squashfs", binaryPath), nmstateDir)
+	if err != nil {
+		log.Errorf("failed to unsquashfs root.squashfs: %v", err.Error())
+		return "", err
+	}
+	return binaryPath, nil
+}
+
+//go:generate mockgen -package=isoeditor -destination=mock_executer.go . Executer
+type Executer interface {
```
rccrdpccl, line 115: nit: I believe the correct spelling would be `Executor`

pkg/isoeditor/nmstate_handler.go
```diff
@@ -0,0 +1,134 @@
...
+		if removeErr != nil {
+			log.WithError(removeErr).Error("failed to remove nmstate temp dir")
+		}
+	}()
+
+	return err
+}
+
+func (n *nmstateHandler) extractNmstatectl(rootfsPath, nmstateDir string) (string, error) {
+	_, err := n.executer.Execute(fmt.Sprintf("7z x %s", rootfsPath), nmstateDir)
```
rccrdpccl, line 82: Any reason we picked system's 7z over native golang bindings?

Dockerfile.image-service-build
```diff
@@ -8,13 +8,17 @@ RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/i
 
 FROM quay.io/centos/centos:stream9
 
+RUN dnf install -y epel-release
```
rccrdpccl, line 11: nit: Using one RUN statement instead of two would generate one layer instead of two

rccrdpccl:
> Also I wonder if startup time is affected.
> 
> @rccrdpccl what do you think?

It looks like it's going to be affected. @linoyaslan could you please compare the time taken to generate a minimal ISO with or without nmstate? I understand at present only an handful of images will take longer, but as we'll release more OpenShift versions more and more images will have this. Can we roughly quantify how much longer that is?
I do not think there's much we can do in the short term, apart from taking it into account and change alerts/probes etc

However I do believe we should try to come up with some long-term plan to try reduce the boot time

pkg/isoeditor/nmstate_handler.go
rccrdpccl: Nevermind I might be wrong here, I was just not used to the word "executer" but I think it might be more appropriate

pkg/isoeditor/nmstate_handler.go
linoyaslan: When I tested it, I encountered an error... the explanation according to @jhernand  is that go-diskfs does not support the zstd compression format used by CoreOS.

rccrdpccl: I am not familiar with this area, is this a blocker or something we could get around by using go libraries for zstd in combination with go-diskfs?

jhernand: There is no workaround: squashfs compression is not something that you can plug into go-diskfs without modifying it. When I checked some time ago they had support in the master branch, but not in any release. I did test in that master branch, and it crashed. They may have improved/fixed it in the latest release.

rccrdpccl: Thanks @jhernand for the explanation. Maybe we can leave a TODO note in the code, create a ticket for follow up and address it once the feature is mature enough? I think it's important we try to avoid shelling out processes like this if we can. WDYT?

pkg/isoeditor/nmstate_handler.go
linoyaslan: I faced an issue when trying to decompress the rootfs due to a format problem that only worked with 7z. Together with Daniel, I tried several Go libraries, but none of them worked for us..

linoyaslan:
@rccrdpccl 

Assisted-image pod startup time:
  - Latest image: 6m30s
  - With new changes: 6m40s

- nmstate handler execution time: ~1 second

Minimal ISOs build time:
  - New changes: 53.687796663s
  - Master branch: 40.32088459s

rccrdpccl:
> @rccrdpccl
> 
> Assisted-image pod startup time:
> 
>     * Latest image: 6m30s
> 
>     * With new changes: 6m40s
> 
> 
> nmstate handler execution time: ~1 second

Would this mean one second more per minimal ISO (of the right version)? Was this tested [with CPU limit](https://github.com/openshift/assisted-image-service/blob/main/deploy/template.yaml#L218)? 

> Minimal ISOs build time:
> 
>     * New changes: 53.687796663s
> 
>     * Master branch: 40.32088459s

We should probably add a `startupProbe` , tweak resources/limits and alerts to mitigate this, but I understand this PR is just another drop that increases boot time, maybe it'd be worth creating a new ticket for that - WDYT @carbonin ?

carbonin:
Do we not already have a ticket to decrease the boot time of the image service? Not sure we need another one specifically for this.

I mostly wanted to make sure this wouldn't push us over some edge of either storage space or time.

rccrdpccl:
> Do we not already have a ticket to decrease the boot time of the image service? Not sure we need another one specifically for this.

Not sure we have a ticket for boot time, I've created a [new one](https://issues.redhat.com/browse/MGMT-19211) now where we can track progress on this. What I was suggesting is to take measures to mitigate increase of boot time while we'll reduce boot time in the longer term (startupProbe and tweaking alerts so we won't get so many false positives).
I've created a different [issue](https://issues.redhat.com/browse/MGMT-19210) where we'll take care of mitigating the slow boot time separately, so we can merge this PR without worrying about it.

> 
> I mostly wanted to make sure this wouldn't push us over some edge of either storage space or time.

I think it won't, but it wouldn't hurt to have parametrized probes so we can tweak them in case it does. @linoyaslan would you mind adding some [useful fields for probes](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#probe-v1-core)? I'm thinking mainly success/failure threshold and period, this will make sure if we trigger some condition we can tweak probes and start the service

carbonin:
Based on https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes it seems like we could probably use the same check as we already do for readiness.

deploy/template.yaml
```diff
@@ -221,10 +233,16 @@ objects:
             httpGet:
               path: /live
               port: 8080
+            failureThreshold: ${LIVENESS_PROBE_FAILURE_THRESHOLD}
```
rccrdpccl, line 236: sorry, my bad but I think this should be with double braces `${{LIVENESS_PROBE_FAILURE_THRESHOLD}}` (all of these parameters are int)

Dockerfile.image-service
```diff
@@ -43,7 +43,8 @@ RUN mkdir $DATA_DIR && chmod 775 $DATA_DIR && chown $UID:$GID /data
 VOLUME $DATA_DIR
 ENV DATA_DIR=$DATA_DIR
 
-RUN dnf install -y nmstate && dnf clean all
+RUN dnf install -y epel-release && \
+    dnf install -y p7zip-plugins squashfs-tools && dnf clean all
```
carbonin, line 47: We need to add these in the downstream dockerfile as well (assuming theres a chance MCE doesn't move fully to konflux builds for 2.8). But this can be done after the merge.

carbonin (approved):

pkg/isoeditor/nmstate_handler.go
linoyaslan: @eifrach

internal/common/version.go
```diff
@@ -0,0 +1,17 @@
...
+func VersionGreaterOrEqual(version1, version2 string) (bool, error) {
+	v1, err := version.NewVersion(version1)
+	if err != nil {
+		return false, err
+	}
+	v2, err := version.NewVersion(version2)
+	if err != nil {
+		return false, err
+	}
+	return v1.GreaterThanOrEqual(v2), nil
```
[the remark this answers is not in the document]
rccrdpccl: why rc (Release Candidate) should be greater than released version? I think rc shouldn't be greater than non-rc, am I missing something here?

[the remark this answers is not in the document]
linoyaslan: resolved here - https://github.com/openshift/assisted-image-service/pull/286

PR #288, opened 2024-11-13
MGMT-19271: Using CPIO as an alternative approach to extract the rootfs instead of p7zip

linoyaslan:
Following this thread - https://redhat-internal.slack.com/archives/C02AX10EQJW/p1731502735110969 , since we can't use the epel pkg, specifically the p7zip pkg, we are now using CPIO to extract the root filesystem.

## Assignees

/cc @carbonin 
/cc @danielerez 

## Checklist

### DIFF
--- a/pkg/isoeditor/nmstate_handler.go
+++ b/pkg/isoeditor/nmstate_handler.go
@@ -80,9 +80,9 @@ func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath string) er
 
 // TODO: Update the code to utilize go-diskfs's squashfs instead of unsquashfs once go-diskfs supports the zstd compression format used by CoreOS - MGMT-19227
 func (n *nmstateHandler) extractNmstatectl(rootfsPath, nmstateDir string) (string, error) {
-	_, err := n.executer.Execute(fmt.Sprintf("7z x %s", rootfsPath), nmstateDir)
+	_, err := n.executer.Execute(fmt.Sprintf("cat %s | cpio -i", rootfsPath), nmstateDir)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


danielerez:
Looks good to me. Please update the version in ticket.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -80,9 +80,9 @@ func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath string) er
 
 // TODO: Update the code to utilize go-diskfs's squashfs instead of unsquashfs once go-diskfs supports the zstd compression format used by CoreOS - MGMT-19227
 func (n *nmstateHandler) extractNmstatectl(rootfsPath, nmstateDir string) (string, error) {
-	_, err := n.executer.Execute(fmt.Sprintf("7z x %s", rootfsPath), nmstateDir)
+	_, err := n.executer.Execute(fmt.Sprintf("cat %s | cpio -i", rootfsPath), nmstateDir)
```
carbonin, line 83: Interested in why this works. I guess the rootfs was always just a cpio archive and not compressed at all?

Did `7z` just handle cpio as well?

danielerez: Yeah, it's cpio archive in 4.18:
```
file rootfs.img
rootfs.img: ASCII cpio archive (SVR4 with no CRC)
```
Which 7z also supports:
```
7z i | grep cpio
0      M        Cpio     cpio
```
I've remembered the file was compressed in some older versions though :/

carbonin: >I've remembered the file was compressed in some older versions though :/

Might be good to figure out if/when this was the case.

PR #291, opened 2024-11-27
MGMT-19371: Add downstream dockerfile for mce

giladravid16:
## Description

As part of migrating to konflux, we need to update our downstream dockerfiles for mce and put them in our upstream repos.

## Assignees

/cc @pastequo 

## Links

Part-of [MGMT-19371](https://issues.redhat.com//browse/MGMT-19371)

## Checklist

### DIFF
--- a/Dockerfile.image-service-mce
+++ b/Dockerfile.image-service-mce
@@ -1,0 +1,60 @@
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/go-toolset:1.21 AS builder
+ARG TARGETOS
+ARG TARGETARCH
+
+ENV COMPONENT_NAME=assisted-image-service
+ENV COMPONENT_VERSION=1.0.0
+ENV COMPONENT_TAG_EXTENSION=" "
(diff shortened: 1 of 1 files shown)


Dockerfile.image-service-mce
```diff
@@ -0,0 +1,60 @@
...
+COPY --chown=${USER_UID} . /app
+WORKDIR /app
+
+RUN go install github.com/google/go-licenses@v1.6.0
+RUN ${HOME}/go/bin/go-licenses save --save_path /tmp/licenses ./...
+
+RUN GO111MODULE=on GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -tags strictfipsruntime -o assisted-image-service main.go
+
+
+FROM --platform=$BUILDPLATFORM registry.access.redhat.com/ubi9/ubi:9.4
```
pastequo, line 24: We use to have a mechanism that allow to update this tag (that was not floating)
Indeed this mechanism won't be supported anymore here but we should probably keep the non-floating tag (for build reproducibility) and see if we can use renovate to mimic the following tag mechanism

PR #328, opened 2025-01-29
MGMT-19819: Add the commit reference from which the image is built to the image

danmanor:
## Description
Currently, in most of assisted installer components CI images we don't have a way to tell from which commit reference the image was built. Since We use an image stream for each component, and we import these streams from one CI component configuration to another, we might end up with images to are not up-to-date. In this case, we would like to have the ability to check if this is actually the case.

## How was this code tested?
CI tests image building, also built locally.

## Assignees

/cc @adriengentil 
/cc @pastequo 
/cc @rccrdpccl  

## Links
No Links

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-.git
 build
 data
 **/Dockerfile.*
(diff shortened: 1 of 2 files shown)


CrystalChun (approved):
/lgtm

PR #337, opened 2025-02-14
MGMT-19211: add limited parallelism to image creation phase

zszabo-rh:
[MGMT-19211](https://issues.redhat.com//browse/MGMT-19211): add limited parallelism to image creation phase

Introduced new env variable MINIMAL_ISO_CREATION_WORKERS to control how many images can be be processed in parallel. This is expected to improve the startup time of the pod, while keeping the required storage space in bay. The default is 2.

Image sizes can vary, but as a rough estimation each downloaded image adds ~1.2Gi to the final storage requirement, and an additional temp disk space of the same size while being extracted and processed. So the minimum disk space required for storing and processing the images can be estimated as (N + M) x 1.2Gi, where N is the number of images and M is the number of workers.

So far the tests shown mixed results: 
- the service running in podman container was able to start significantly faster when parallel processing was used
- no measurable improvement could be detected during similar tests in k8s (assisted-test-infra).
- the pods on integration env gained minor improvement in startup times, it took ~10% shorter time to get the API enabled with parallel image processing, regardless of the number of workers (tested with 2 and 5 workers, with proportionally increased resource limits). 

Based on the test results the default value for the number of workers was set to 2.

This change requires all the workdirs to be unique throughout the ISO creation process. `CreateNmstateRamDisk` was the only function that still used static temp folder, so it was also modified to create randomized folders.

/cc @rccrdpccl 
/cc @danielerez 
/cc @carbonin 

## Checklist


deploy/template.yaml
```diff
@@ -170,11 +170,13 @@ parameters:
 - name: CPU_REQUEST
   value: "100m"
 - name: CPU_LIMIT
-  value: "200m"
+  value: "1"
 - name: MEMORY_REQUEST
   value: "400Mi"
 - name: MEMORY_LIMIT
   value: "800Mi"
+- name: IMAGE_WORKER_LIMIT
```
carbonin, line 178: Why make this name different from the env in the service? Pick the better one and use it everywhere.

pkg/imagestore/imagestore.go
```diff
@@ -276,6 +276,10 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 		return err
 	}
 
+	const maxWorkers = 3
```
carbonin, line 279: It doesn't look like this is used, can it be removed?

deploy/template.yaml
carbonin: That said, maybe we should make both of these specific to the minimal image creation in some way since we still allow unbounded parallelism for the full image download (which could also be considered an image worker an cause confusion)

pkg/isoeditor/nmstate_handler.go
```diff
@@ -30,14 +30,19 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
...
 	var err error
-	nmstateDir := filepath.Join(n.workDir, "nmstate")
-	err = os.MkdirAll(nmstateDir, os.ModePerm)
-	if err != nil {
-		return err
+
+	// Prepare working folder for nmstate
+	if nmstateDir == "" {
+
+		nmstateDir, err = os.MkdirTemp(n.workDir, "nmstate")
```
carbonin, line 39: Can you add a bit of explanation to the PR and commit explaining why this is needed?

I think I understand how this is related to the addition of parallelism, but we should write it down.

pkg/isoeditor/nmstate_handler_test.go
```diff
@@ -67,7 +67,7 @@ var _ = Context("with test files", func() {
 		})
 
 		It("ram disk created successfully", func() {
-			err := nmstateHandler.CreateNmstateRamDisk("", ramDiskPath)
+			err := nmstateHandler.CreateNmstateRamDisk("", ramDiskPath, filepath.Join(extractDir, "nmstate"))
```
carbonin, line 70: I'm not sure that the new parameter is needed here since it's really only used for testing and in this case `extractDir` is `os.TempDir` which is also what is being passed to `NewNmstateHandler` as `workDir`

carbonin (changes requested):

deploy/template.yaml
zszabo-rh: Changed it to IMAGE_WORKERS. Or are you suggesting something more descriptive name?

pkg/isoeditor/nmstate_handler.go
zszabo-rh: I had to inject fixed nmstate folder name to the function to make unit test pass. The test suite prepares some dummy files here before running the tc, which was incompatible with the new randomized folders concept. This is a bit hackish, but I didn't have a better idea how to overcome the failing unit test issue.

pkg/isoeditor/nmstate_handler_test.go
zszabo-rh: I think this is needed, this is the part where I pass the prepared nmstate folder path to the function.

rccrdpccl: It looks like it's a directory only used for temporary purposes, would it work if we use a unique dir (os.TempDir for instance) inside CreateNmstateRamDisk function, or am I missing something?

pkg/imagestore/imagestore.go
```diff
@@ -287,25 +290,31 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
+			minimalPath := filepath.Join(s.dataDir, isoFileName(ImageTypeMinimal, openshiftVersion, imageVersion, arch))
+			if _, err := os.Stat(minimalPath); os.IsNotExist(err) {
+				log.Infof("Creating minimal iso for %s-%s-%s", openshiftVersion, imageVersion, arch)
 
-			fullPath := filepath.Join(s.dataDir, isoFileName(ImageTypeFull, openshiftVersion, imageVersion, arch))
-			rootfsURL, err := buildRootfsURL(s.imageServiceBaseURL, arch, openshiftVersion)
-			if err != nil {
-				return fmt.Errorf("failed to build rootfs URL: %v", err)
-			}
+				fullPath := filepath.Join(s.dataDir, isoFileName(ImageTypeFull, openshiftVersion, imageVersion, arch))
```
rccrdpccl, line 298: Not relevant to this PR, but it looks like we're processing everything on the PV. I am not familiar with the internals of building the image: is the process more I/O intensive than just copying something from memory to the disk? Asking because if it is, we might benefit from working in tmpfs (and maybe to some extent also in ephemeral storage) and copying the final result. @carbonin thoughts?

pkg/isoeditor/nmstate_handler.go
carbonin: I wasn't talking about the unit test issue. I was more referencing why you needed to touch this file (`nmstate_handler.go`) at all.

I just think that warrants a callout in the commit/PR description

rccrdpccl:
/lgtm
/hold

holding to give other people a chance to review

zszabo-rh:
obsolated by https://github.com/openshift/assisted-image-service/pull/349

PR #340, opened 2025-02-17
NO-ISSUE: Add `dump.erofs` to the image

jhernand:
## Description

This patch adds the `dump.erofs` binary to the image. To do so it fetches the source and builds it inside a separate _tools_ container.


jhernand:
/hold

This is a proof of concept, don't merge it.

packages/README.md
```diff
@@ -0,0 +1,4 @@
+# External packages
+
+This directory contains the source code of packages that we need, but aren't available in the
+distribution that we use, so we build them from source.
```
carbonin, line 4: Can we list where we get these from for when we need to upgrade?

Dockerfile.image-service-mce
```diff
@@ -32,6 +84,7 @@ ENV DATA_DIR=$DATA_DIR
 
 RUN dnf -y update && dnf install -y cpio squashfs-tools && dnf clean all
 
+COPY --from=tools /usr/local /usr/local
```
carbonin, line 87: Should/Can we just copy what we need to ensure we're not overwriting anything here?

packages/README.md
jhernand: I moved the code to build the packages to separate `build.sh` scripts inside the `packages` directory. For example, for `xz` it is now in `packages/xz/bulid.sh`. Those scripts contain a `version` environment variable that can be used to update the package. I tried to document it in the new version of this README.md.

Dockerfile.image-service-mce
jhernand: Yes, modified this to copy the very minimum that we need: the `dump.erofs` binary and the `liblzma.so.5` shared library.

linoyaslan:
Looks good to me, but let's wait for Nick to see if he has any other comments

carbonin (approved):

PR #355, opened 2025-03-04
OCPBUGS-51209:  Fixing artifact name issue for vmlinuz in  latest rhelVersions

veera-damisetti:
Fixing artifact name issue for vmlinuz in  latest rhelVersions ( starting from 9.6 ) 

- Updated boot_artifacts.go to change artifact name to vmliuz from kernel.img based on RHCOS version

## Description

-  Starting from RHEL9.6 , artifact name for kernel has  changed to vmliuz from kernel.img. this is breaking the existing flow in Hosted Control Plane and Assisted Installer.

- Updated the logic to change the artifact name based the RHEL Version. 

## How was this code tested?

- Built a test image with assisted-image-service 
- Using this test image, created AgentServiceConfig and InfraEnv objects, and validated the images generated in InfraEnv.

## Assignees

@AmadeusPodvratnik 

cc @v78singh 
cc @phani2898 

## Links
https://issues.redhat.com/browse/OCPBUGS-51209

## Checklist

### DIFF
--- a/internal/handlers/boot_artifacts.go
+++ b/internal/handlers/boot_artifacts.go
@@ -6,7 +6,6 @@ import (
 	"net/url"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/openshift/assisted-image-service/pkg/imagestore"
@@ -22,38 +21,18 @@ var _ http.Handler = &BootArtifactsHandler{}
 
 var bootpathRegexp = regexp.MustCompile(`^/boot-artifacts/(.+)`)
 
-func parseArtifact(path, arch, version, isoFileName string) (string, error) {
+func parseArtifact(path, arch string) (string, error) {
 	match := bootpathRegexp.FindStringSubmatch(path)
 	if len(match) < 1 {
 		return "", fmt.Errorf("malformed download path: %s", path)
 	}
 
 	var artifact string
-	var rhelVersion int
-	var err error
-
-	// Fetching rhcosVersion from isoFileName
-
-	rhcosVersionSplit := strings.Split(isoFileName, version+"-")
-	rhcosVersion := rhcosVersionSplit[len(rhcosVersionSplit)-1]
-
-	// Fetching rhelVersion from rhcosVersiom
-	if len(strings.Split(rhcosVersion, ".")) > 1 {
-		rhelVersion, err = strconv.Atoi(strings.Split(rhcosVersion, ".")[1])
-	}
-
-	if err != nil {
-		fmt.Println("Error in fetching RHCOS Version from ISO file")
-		return "", err
-	}
 	switch match[1] {
 	case "rootfs":
 		artifact = "rootfs.img"
 	case "kernel":
 		artifact = "vmlinuz"
-		if arch == "s390x" && rhelVersion < 96 {
-			artifact = "kernel.img"
-		}
 	case "ins-file":
 		if arch == "s390x" {
 			artifact = "generic.ins"
@@ -64,6 +43,15 @@ func parseArtifact(path, arch, version, isoFileName string) (string, error) {
 		return "", fmt.Errorf("malformed download path: %s", path)
 	}
 	return artifact, nil
+}
+
+func getArtifactFilePath(artifact string) string {
+	filePath := fmt.Sprintf("/images/pxeboot/%s", artifact)
+	if artifact == "generic.ins" {
+		// s390x only, unlike other artifacts this one is at the root of the ISO
+		filePath = fmt.Sprintf("/%s", artifact)
+	}
+	return filePath
 }
 
 func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


internal/handlers/boot_artifacts.go
```diff
@@ -19,23 +20,29 @@ type BootArtifactsHandler struct {
 
 var _ http.Handler = &BootArtifactsHandler{}
 
+var IsoFileName string
```
carbonin, line 23: Don't create or use globals like this.

carbonin (changes requested):
This should not be done based on a file name.
Users can mirror these ISOs and change the name to whatever they want.

veera-damisetti:
> This should not be done based on a file name. Users can mirror these ISOs and change the name to whatever they want.

Thanks @carbonin  for the quick review. 

Even if we are mirroring ISO and changing the name while providing ISO URL, **_IsoFileName_** is always coming with version and RHCOS version.

For example, we have ISO URL from mirror, https://mirror.openshift.com/pub/openshift-v4/s390x/dependencies/rhcos/4.18/4.18.1/rhcos-4.18.1-s390x-live.s390x.iso 
Still IsoFilename is getting generated with version and RHCOS version. 
`
Printing ISO File name for debug: /data/rhcos-full-iso-4.18.1-418.94.202502100215-0-s390x.iso`

Is this the scenario which you are indicating where user can mirror ISO and change the name ?

carbonin:
> Still IsoFilename is getting generated with version and RHCOS version

Okay, I remember this now. Let me take a closer look when I get time and see if this still makes sense.

carbonin:
>Still IsoFilename is getting generated with version and RHCOS version.
 Printing ISO File name for debug: /data/rhcos-full-iso-4.18.1-418.94.202502100215-0-s390x.iso
 
Okay, so the issue here is still that these values (`4.18` and `418.94.202502100215-0`) are still coming from user input and could technically be arbitrary identifiers.
I'm not sure we actually validate their format or correctness anywhere.
Unfortunately we're already relying on these being valid to some degree with the nmstate changes @linoyaslan made, but maybe we want to re-evaluate that and use the actual ISO volume identifier for a better source of truth for these kinds of versions.

In any case I think a simpler solution for you would be to just look for both files in the ISO rather than trying to identify which file is the correct one based on the version.

I'm not sure if this is 100% correct as I put it together rather quickly, but maybe something like this instead:
```patch
diff --git a/internal/handlers/boot_artifacts.go b/internal/handlers/boot_artifacts.go
index 0f09b9c..7e4fc20 100644
--- a/internal/handlers/boot_artifacts.go
+++ b/internal/handlers/boot_artifacts.go
@@ -49,6 +49,15 @@ func parseArtifact(path, arch string) (string, error) {
        return artifact, nil
 }
 
+func getArtifactFilePath(artifact string) string {
+       filePath := fmt.Sprintf("/images/pxeboot/%s", artifact)
+       if artifact == "generic.ins" {
+               // s390x only, unlike other artifacts this one is at the root of the ISO
+               filePath = fmt.Sprintf("/%s", artifact)
+       }
+       return filePath
+}
+
 func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
        if r.Method != http.MethodGet && r.Method != http.MethodHead {
                log.Error("Only GET and HEAD methods are supported with this endpoint.")
@@ -70,13 +79,11 @@ func (b *BootArtifactsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
        }
 
        isoFileName := b.ImageStore.PathForParams(imagestore.ImageTypeFull, version, arch)
-       file_path := fmt.Sprintf("/images/pxeboot/%s", artifact)
-       if artifact == "generic.ins" {
-               // s390x only, unlike other artifacts this one is at the root of the ISO
-               file_path = fmt.Sprintf("/%s", artifact)
+       fileReader, err := isoeditor.GetFileFromISO(isoFileName, getArtifactFilePath(artifact))
+       // retry with the old kernel file name if vmlinuz is not found on s390x
+       if err != nil && arch == "s390x" && artifact == "vmlinuz" {
+               fileReader, err = isoeditor.GetFileFromISO(isoFileName, getArtifactFilePath("kernel.img"))
        }
-
-       fileReader, err := isoeditor.GetFileFromISO(isoFileName, file_path)
        if err != nil {
                httpErrorf(w, http.StatusInternalServerError, "Error creating file reader stream: %v", err)
                return
```

Maybe we'd also want to put these artifacts names in constants or something, but I'm just posting this to help you understand my idea for moving this away form relying on parsing out versions in any real way.

veera-damisetti:
Thanks @carbonin for the inputs .
Checking for both  files in the ISO rather than relying on rhcosVersion is a great idea.
I have changed the logic to remove the dependency of isoFileName. Please have a look 
Thanks ..!

carbonin:
The changes look good to me, just need to get tests passing.

veera-damisetti:
Thanks @carbonin  for the update. 

`ci/prow/edge-test` is failing because of come change in artifact name in mirror.
for latest builds, it is coming as rhcos-live-iso.aarch64.iso, but in code URL it is rhcos-live.aarch64.iso


`"failed to download https://mirror.openshift.com/pub/openshift-v4/arm64/dependencies/rhcos/pre-release/latest/rhcos-live.aarch64.iso: request to https://mirror.openshift.com/pub/openshift-v4/arm64/dependencies/rhcos/pre-release/latest/rhcos-live.aarch64.iso returned error code 404",`


Will try to ask about this in channel
thanks

veera-damisetti:
@carbonin unit tests check is now passing after getting fix from [PR]( https://github.com/openshift/assisted-image-service/pull/359/files)
Can you please have look once

pastequo:
@veera-damisetti I think you have to rebase your branch. Your local branch contains old prefetch rpm requirement and tekton is using the local branch to build

veera-damisetti:
Thanks @rccrdpccl 
I did rebase and checks are passing now 👍🏼 , please have look once 
cc: @carbonin

veera-damisetti:
@carbonin Can you please a look and give lgtm , if everything is fine
thanks ..!

veera-damisetti:
Thanks @carbonin  @rccrdpccl 
 for quick review and inputs.

PR #357, opened 2025-03-12
NO-ISSUE: Fix Konflux configuration for ocm-2.14

danmanor:


danmanor:
@giladravid16 any idea why this one fail ?

danmanor:
retested several times and I don't see error in the diff

danmanor:
@giladravid16 do you know a way to solve it ?

giladravid16:
We can try to regenerate the rpm lock file in the rpm-prefetching dir and see if we get a different file.
The error message in the build also suggested adding --skip-broken.

PR #400, opened 2025-05-07
MGMT-20597: Add nmstatectl to initrd for PXE provisioning

linoyaslan:
## Description
When booting via PXE with static networking, connectivity could not be established, which resulted in failure to fetch the rootfs. The root cause is the absence of nmstatectl in the initrd. This PR solve this.

**Notes:** 
1. This PR triggered a high-severity warning from Snyk related to os.Open. Although the suggested mitigation was applied in the PR, the warning still persists. That said, it doesn’t appear to be a real security risk because:

-  The input parameters (version and arch) are validated.
- Even if someone tried to craft values to access files outside the intended directory, an additional check was added to ensure that the parent directory of nmstatectlPath matches the expected working directory - and the code rejects the path if it doesn’t.

2. The reason we're placing the nmstatectl cache file creation as part of the minimal ISO creation - despite it not being directly related - is because in that flow we already have access to the rootfs URL and extract the nmstatectl binary there. While it may not be the most logically accurate place, it's the more practical choice, as creating a separate caching flow just for PXE would be redundant.

## How was this code tested?

1. Select static networking when creating the cluster.
3. Generate the PXE script.
4. Extract the initrd and kernel URLs from the PXE script and download the images.
5. Boot from the artifacts (initrd.img and kernel.img) using the kargs provided in the PXE script.
6. Disable DHCP by adding the following karg: ip=::::<interface-name>:none
7. Start the boot process

## Assignees

/cc @danielerez  
/cc @carbonin 

## Checklist

### DIFF
--- a/internal/handlers/initrd.go
+++ b/internal/handlers/initrd.go
@@ -4,7 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"os"
 	"time"
+
+	"github.com/openshift/assisted-image-service/internal/common"
 
 	"github.com/go-chi/chi/v5"
 	log "github.com/sirupsen/logrus"
@@ -89,6 +92,25 @@ func initrdOverlayReader(imageStore imagestore.ImageStore, client *AssistedServi
 			return nil, "", http.StatusInternalServerError, fmt.Errorf("failed to create append reader for initrd: %v", err)
 		}
 
+		versionOK, err := common.VersionGreaterOrEqual(version, isoeditor.MinimalVersionForNmstatectl)
+		if err != nil {
+			return nil, "", http.StatusInternalServerError, err
+		}
+
+		if versionOK {
+			nmstatectlPath, err := imageStore.NmstatectlPathForParams(version, arch)
+			if err != nil {
+				return nil, "", http.StatusInternalServerError, err
+			}
+			nmstateImgContent, err := os.Open(nmstatectlPath)
+			if err != nil {
+				return nil, "", http.StatusInternalServerError, fmt.Errorf("failed to read nmstate img: %v", err)
+			}
+			initrdReader, err = overlay.NewAppendReader(initrdReader, nmstateImgContent)
+			if err != nil {
+				return nil, "", http.StatusInternalServerError, fmt.Errorf("failed to create append reader for initrd: %v", err)
+			}
+		}
 	}
 
 	return initrdReader, lastModified, 0, nil
--- a/pkg/isoeditor/nmstate_handler.go
+++ b/pkg/isoeditor/nmstate_handler.go
@@ -15,7 +15,7 @@ import (
 
 //go:generate mockgen -package=isoeditor -destination=mock_nmstate_handler.go . NmstateHandler
 type NmstateHandler interface {
-	CreateNmstateRamDisk(rootfsPath, ramDiskPath string) error
+	CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatectlPath string) error
 }
 
 type nmstateHandler struct {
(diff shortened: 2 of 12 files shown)


internal/handlers/initrd.go
```diff
@@ -89,6 +95,30 @@ func initrdOverlayReader(imageStore imagestore.ImageStore, client *AssistedServi
...
+			extractDir, err := os.MkdirTemp(workDir, "isoutil")
+			if err != nil {
+				return nil, "", http.StatusInternalServerError, err
+			}
+			defer os.RemoveAll(extractDir)
+			if err = isoeditor.Extract(isoPath, extractDir); err != nil {
+				return nil, "", http.StatusInternalServerError, err
+			}
+			rootfsPath := filepath.Join(extractDir, "images/pxeboot/rootfs.img")
+			nmstateCpioArchive, err := nmstateHandler.BuildNmstateCpioArchive(rootfsPath)
```
danielerez, line 113: So IIUC, it means that we're extracting the binary and building the archive for every '/pxe-initrd' request.
I.e. we probably need to consider caching of the binary. Maybe we could simply save the cpis in the ImageStore when handling the templates? It would also simply the logic here. Instead of rebuilding, just read it from the ImageStore, e.g. something like: https://github.com/openshift/assisted-image-service/blob/1362d5b6f4e729ab797704bd5c48535dc2192a4e/internal/handlers/initrd.go#L75

carbonin: Yeah I wonder what would happen if we got a ton of those requests at the same time.
Probably nothing good.

internal/handlers/initrd.go
```diff
@@ -89,6 +92,22 @@ func initrdOverlayReader(imageStore imagestore.ImageStore, client *AssistedServi
...
+			return nil, "", http.StatusInternalServerError, err
+		}
+
+		if versionOK {
+			nmstatectlPath := imageStore.NmstatectlPathForParams(version, arch)
+			nmstateImgContent, err := os.ReadFile(nmstatectlPath)
+			if err != nil {
+				return nil, "", http.StatusInternalServerError, fmt.Errorf("failed to read nmstate img: %v", err)
+			}
+			initrdReader, err = overlay.NewAppendReader(initrdReader, bytes.NewReader(nmstateImgContent))
```
carbonin, line 106: There's no reason to read the content into memory then create a reader for the content.
`os.Open(nmstatectlPath)` should give you a reader that can then be passed to `NewAppendReader`

internal/handlers/initrd.go
```diff
@@ -89,6 +93,25 @@ func initrdOverlayReader(imageStore imagestore.ImageStore, client *AssistedServi
...
 		}
 
+		versionOK, err := common.VersionGreaterOrEqual(version, isoeditor.MinimalVersionForNmstatectl)
+		if err != nil {
+			return nil, "", http.StatusInternalServerError, err
+		}
+
+		if versionOK {
+			nmstatectlPath := imageStore.NmstatectlPathForParams(version, arch)
+			if filepath.Dir(nmstatectlPath) != imageStore.GetDataDir() {
```
carbonin, line 103: Instead of creating this getter, how about we return an error from `NmstatectlPathForParams` if the path ends up escaping the data directory?

pkg/isoeditor/nmstate_handler.go
```diff
@@ -30,52 +30,67 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
...
+		return err
+	}
+
+	// write for RAM disk
+	err = os.WriteFile(ramDiskPath, compressedCpio, 0755) //nolint:gosec
+	if err != nil {
+		return err
+	}
+
+	// write for caching
```
carbonin, line 45: Can you also add to the commit and PR description the reasoning behind why we're doing this with the minimal ISO creation even though caching this has nothing to do with the minimal ISO?

carbonin:
/lgtm

Thanks for the patience on this one @linoyaslan

PR #445, opened 2025-07-08
MGMT-20830: Support for serving an OVE image (ISO) with an embedded custom ignition

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-20830?filter=-1

### DIFF
--- a/internal/handlers/assisted_service_client.go
+++ b/internal/handlers/assisted_service_client.go
@@ -95,7 +95,7 @@ func (c *AssistedServiceClient) ramdiskContent(imageServiceRequest *http.Request
 // ignitionContent returns the ignition data on success and the error and the corresponding http status code
 // The code is also returned to ensure issues with authentication from the assisted service request are communicated back to the image service user
 // The returned code should only be used if an error is also returned
-func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string, imageType string) (*isoeditor.IgnitionContent, string, int, error) {
+func (c *AssistedServiceClient) ignitionContent(imageServiceRequest *http.Request, imageID string, imageType string, isOVE bool) (*isoeditor.IgnitionContent, string, int, error) {
 
 	u := url.URL{
 		Scheme: c.assistedServiceScheme,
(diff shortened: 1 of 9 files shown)


carbonin:
Is there a way we can detect the type of the ISO so the user won't have to explicitly tell us?
Once we download it maybe there is some kind of marker file we can look at?

danielerez (commented):
For downloading an OVE image, we need to support the new ImageType in assisted-image-service.
I.e. something like:
* In imagestore -> Populate [func](https://github.com/openshift/assisted-image-service/blob/63d1c4c78c41ee965b1ee3a0b6ab25ed2a5a7c03/pkg/imagestore/imagestore.go#L237), read the image type:
```
imageType := imageInfo["type"]
```
then, generate the [fullPath](https://github.com/openshift/assisted-image-service/blob/63d1c4c78c41ee965b1ee3a0b6ab25ed2a5a7c03/pkg/imagestore/imagestore.go#L251) accordingly.
* Add the new image type to parseShortURL [func](https://github.com/openshift/assisted-image-service/blob/63d1c4c78c41ee965b1ee3a0b6ab25ed2a5a7c03/internal/handlers/short.go#L49).
E.g.
```
	case "disconnected.iso":
		params.imageType = "disconnected-iso"
```

danielerez:
> Is there a way we can detect the type of the ISO so the user won't have to explicitly tell us? Once we download it maybe there is some kind of marker file we can look at?

Do you mean instead of supporting a new "type" property to the [OsImage](https://github.com/openshift/assisted-image-service/blob/63d1c4c78c41ee965b1ee3a0b6ab25ed2a5a7c03/pkg/imagestore/imagestore.go#L24)?
I indeed wasn't sure about it. Explicitly adding the type explicitly is simple enough when storing the image, and we can stream it according to the path, i.e. as currently do with [PathForParams](https://github.com/openshift/assisted-image-service/blob/63d1c4c78c41ee965b1ee3a0b6ab25ed2a5a7c03/pkg/imagestore/imagestore.go#L312).
But do you think it should be implicit instead? E.g. read the ISO post caching and detect some indication for the image type? Worth the effort?

PR #449, opened 2025-07-09
MGMT-20830: Support for serving an OVE image (ISO) with an embedded custom ignition

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-20830


internal/handlers/short.go
```diff
@@ -51,6 +51,8 @@ func parseShortURL(r *http.Request) (*imageDownloadParams, int, error) {
 		params.imageType = "minimal-iso"
 	case "full.iso":
 		params.imageType = "full-iso"
+	case "disconnected-interactive.iso":
+		params.imageType = "disconnected-interactive-iso"
```
carbonin, line 55: This is the best name we have for this? It's fine if there are no other options, but it feels a bit unwieldy to me.

carbonin (commented):
Do we actually have any of these (disconnected-interactive isos) yet to test with?

internal/handlers/short.go
danielerez: Yeah, naming is hard:) Maybe simply 'disconnected.iso'? @andfasano wdyt?

andfasano: In this case I feel that the `interactive` part plays quite a relevant role, since it's a special ISO meant not only to support the disconnected installation (ie like the regular ABI ISO), but also to spin up a local agent TUI and moreover a complete assisted UI to support the interactive installation process by the user. Anyhow I don't want to block on this, if no other better ideas will come up I'm fine with `disconnected.iso`

andfasano: That's right, `OVE` is only the current scope, so it will be better to avoid using it in the naming for a more generic product

zaneb: In the context of assisted-service (where users will encounter this), all installs are interactive

- `disconnected-iso`
- `self-contained-iso`
- `jumbo-iso`

not gonna use the 'a' word :speak_no_evil:

carbonin: I like just removing interactive.

+1 for `disconnected-iso`/`disconnected.iso` from me

pkg/imagestore/imagestore.go
```diff
@@ -322,6 +338,20 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
 }
 
+func (*rhcosStore) getImageType(imageInfo map[string]string) (string, error) {
+	validTypes := map[string]bool{
+		ImageTypeFull:            true,
+		ImageTypeDisconnectedIso: true,
+	}
+
+	imageType, ok := imageInfo["type"]
+	if !ok || !validTypes[imageType] {
```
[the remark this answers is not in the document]
carbonin: Otherwise we'll break existing deployments when they upgrade

carbonin (commented):
Code looks good, but we should have tests for the new behavior in imagestore.go

pastequo:
Konflux PR has been merged, if you rebase this PR, the ecp check should pass

PR #462, opened 2025-07-24
MGMT-21276: Add nmstatectl binary for the s390x arch

linoyaslan:
Due to [this]( https://github.com/openshift/assisted-image-service/blob/1b29aba9d5bee83a536e40233a8170dbb2c72520/pkg/imagestore/imagestore.go#L286-L290) condition, the nmstatectl binary is never fetched for the s390x architecture.

This fix addresses the issue by decoupling the extraction of nmstatectl from the minimal ISO flow - since minimal ISO is not supported for s390x. Instead, the binary is now cached, ensuring it is available for s390x as well.

Assisted-by: Cursor

## How was this code tested?
Test 1:
1. Created a cluster with the s390x architecture and static network configuration.
2. Downloaded the PXE boot script.
3. Retrieved the initrd image referenced in the script.
4. Verified that nmstatectl is included in the image using binwalk.

Test 2:
1. Created a cluster with the s390x architecture and static network configuration.
2. Downloaded the PXE boot script.
3. Retrieved the initrd, kernel and kernel args
4. Launch a VM on a Beaker s390x machine and boot it using the artifacts.
5. Confirm that the VM boots successfully and is discovered as expected.

## Assignees

/cc @carbonin 
/cc @jhernand 
/cc @danielerez 

## Checklist

### DIFF
--- a/integration_test/images_test.go
+++ b/integration_test/images_test.go
@@ -19,8 +19,10 @@ import (
 
 	"github.com/cavaliercoder/go-cpio"
 	diskfs "github.com/diskfs/go-diskfs"
+	"github.com/diskfs/go-diskfs/filesystem"
 	"github.com/google/uuid"
 	"github.com/onsi/gomega/ghttp"
+	"github.com/openshift/assisted-image-service/internal/common"
 	"github.com/openshift/assisted-image-service/internal/handlers"
 	"github.com/openshift/assisted-image-service/pkg/imagestore"
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
@@ -76,6 +78,18 @@ var (
 			"cpu_architecture":  "x86_64",
 			"url":               "https://okd-scos.s3.amazonaws.com/okd-scos/builds/413.9.202302280609-0/x86_64/scos-413.9.202302280609-0-live.x86_64.iso",
 			"version":           "x86_64-latest",
+		},
+		{
+			"openshift_version": "4.18",
+			"cpu_architecture":  "s390x",
+			"url":               "https://mirror.openshift.com/pub/openshift-v4/s390x/dependencies/rhcos/4.18/4.18.1/rhcos-4.18.1-s390x-live.s390x.iso",
+			"version":           "s390x-418",
+		},
+		{
+			"openshift_version": "4.18",
+			"cpu_architecture":  "x86_64",
+			"url":               "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/4.18/4.18.1/rhcos-4.18.1-x86_64-live.x86_64.iso",
+			"version":           "x86_64-418",
 		},
 	}
 
@@ -235,20 +249,15 @@ var _ = Describe("Image integration tests", func() {
 					Expect(err).NotTo(HaveOccurred())
 
 					By("verifying ignition content")
-					f, err := fs.OpenFile("/images/ignition.img", os.O_RDONLY)
+					rc, err := ignitionPayloadReader(fs, version)
 					Expect(err).NotTo(HaveOccurred())
-					gzipReader, err := gzip.NewReader(f)
-					Expect(err).NotTo(HaveOccurred())
-					cpioReader := cpio.NewReader(gzipReader)
-					hdr, err := cpioReader.Next()
-					Expect(err).NotTo(HaveOccurred())
-					Expect(hdr.Name).To(Equal("config.ign"))
-					Expect(hdr.Size).To(Equal(int64(len(tc.expectedIgnition))))
-					content, err := io.ReadAll(cpioReader)
-					Expect(err).NotTo(HaveOccurred())
-					Expect(content).To(Equal(tc.expectedIgnition))
+					defer rc.Close()
 
-					if tc.expectedRamdisk != nil {
+					got, err := readIgnitionContentFromGzCpio(rc)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(got).To(Equal(tc.expectedIgnition))
+
+					if len(tc.expectedRamdisk) > 0 {
 						By("verifying ramdisk content")
 						f, err := fs.OpenFile("/images/assisted_installer_custom.img", os.O_RDONLY)
 						Expect(err).NotTo(HaveOccurred())
@@ -272,6 +281,82 @@ var _ = Describe("Image integration tests", func() {
 				})
 			}
 		})
+		Context("nmstate archive verification - "+tc.name, func() {
+			BeforeEach(func() {
+				imageID = uuid.New().String()
+
+				// Set up assisted service
+				assistedServer = ghttp.NewServer()
+				u, err := url.Parse(assistedServer.URL())
+				Expect(err).NotTo(HaveOccurred())
+
+				// pxe-initrd path: ignition without discovery_iso_type
+				assistedServer.AppendHandlers(
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/files", imageID), "file_name=discovery.ign"),
+						ghttp.RespondWith(http.StatusOK, tc.expectedIgnition),
+					),
+				)
+				// pxe-initrd path: minimal-initrd always requested; 200 when provided, else 204
+				if len(tc.expectedRamdisk) > 0 {
+					assistedServer.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/minimal-initrd", imageID)),
+							ghttp.RespondWith(http.StatusOK, tc.expectedRamdisk),
+						),
+					)
+				} else {
+					assistedServer.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/minimal-initrd", imageID)),
+							ghttp.RespondWith(http.StatusNoContent, nil),
+						),
+					)
+				}
+
+				asc, err := handlers.NewAssistedServiceClient(u.Scheme, u.Host, "")
+				Expect(err).NotTo(HaveOccurred())
+
+				mdw := middleware.New(middleware.Config{})
+				imageServer = httptest.NewServer(handlers.NewImageHandler(imageStore, asc, 1, mdw))
+				imageClient = imageServer.Client()
+			})
+			AfterEach(func() {
+				assistedServer.Close()
+				imageServer.Close()
+			})
+
+			for i := range versions {
+				version := versions[i]
+
+				It("includes nmstate for "+version["openshift_version"]+" "+version["cpu_architecture"], func() {
+					ok, err := common.VersionGreaterOrEqual(version["openshift_version"], isoeditor.MinimalVersionForNmstatectl)
+					Expect(err).NotTo(HaveOccurred())
+
+					if len(tc.expectedRamdisk) <= 0 || !ok {
+						Skip(fmt.Sprintf("skipping test %s due to ocp version < 4.18 or ramdisk isn't present", tc.name))
+					}
+
+					path := fmt.Sprintf("/images/%s/pxe-initrd?version=%s&arch=%s",
+						imageID, version["openshift_version"], version["cpu_architecture"])
+					resp2, err := imageClient.Get(imageServer.URL + path)
+					Expect(err).NotTo(HaveOccurred())
+					defer resp2.Body.Close()
+					Expect(resp2.StatusCode).To(Equal(http.StatusOK))
+
+					initrdBytes, err := io.ReadAll(resp2.Body)
+					Expect(err).NotTo(HaveOccurred())
+
+					nmPath, err := imageStore.NmstatectlPathForParams(version["openshift_version"], version["cpu_architecture"])
+					Expect(err).NotTo(HaveOccurred())
+					nmBytes, err := os.ReadFile(nmPath)
+					Expect(err).NotTo(HaveOccurred())
+
+					// nmstate archive is appended last when ramdisk is present
+					Expect(bytes.HasSuffix(initrdBytes, nmBytes)).To(BeTrue(), "initrd should end with nmstate archive")
+				})
+			}
+		})
 	}
 })
 
@@ -281,7 +366,8 @@ var _ = BeforeSuite(func() {
 	imageDir, err = os.MkdirTemp("", "imagesTest")
 	Expect(err).To(BeNil())
 
-	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir, isoeditor.NewNmstateHandler(imageDir, &isoeditor.CommonExecuter{})), imageDir, imageServiceBaseURL, false, versions, "", map[string]string{}, map[string]string{})
+	nmstateHandler := isoeditor.NewNmstateHandler(imageDir, &isoeditor.CommonExecuter{})
+	imageStore, err = imagestore.NewImageStore(isoeditor.NewEditor(imageDir, nmstateHandler), imageDir, imageServiceBaseURL, false, versions, "", map[string]string{}, map[string]string{}, nmstateHandler)
 	Expect(err).NotTo(HaveOccurred())
 
 	err = imageStore.Populate(context.Background())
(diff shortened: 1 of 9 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -351,6 +416,8 @@ func (s *rhcosStore) cleanDataDir() error {
 	for _, version := range s.versions {
 		// Only add full isos here as we want to regenerate the minimal image on each deploy
 		expectedFiles = append(expectedFiles, isoFileName(ImageTypeFull, version["openshift_version"], version["version"], version["cpu_architecture"]))
+		// Keep nmstatectl cached files for all architectures (including s390x)
+		expectedFiles = append(expectedFiles, nmstatectlFileName(version["openshift_version"], version["version"], version["cpu_architecture"]))
```
linoyaslan, line 420: I'd appreciate input from reviewers, particularly on this point  as I'm unsure whether it's better to keep the nmstatectl after cleanup, or to re-extract it on each deploy
Thank you!

pkg/imagestore/imagestore.go
```diff
@@ -283,6 +284,11 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 		imageVersion := imageInfo["version"]
 		arch := imageInfo["cpu_architecture"]
 
+		// Extract and cache nmstatectl for all architectures
```
danielerez, line 287: I wonder if maybe we should simply remove the s390x filtering in line 294.
I mean, instead of changing the flow, we could just create a minimal ISO also for s390x.
Do we know if it's indeed still not applicable using s390x for minimal ISO?
If so, is it worth complicating the nmstatectl flow just to avoid creating redundant minimal ISOs?
No strong opinion, but something to consider:)

pkg/imagestore/imagestore.go
danielerez: Sounds reasonable to keep it in cache. Mainly as these binaries shouldn't change once extracted from ISO.
So seems redundant to waste resources on this for every deploy.

pkg/imagestore/imagestore.go
linoyaslan: > Do we know if it's indeed still not applicable using s390x for minimal ISO?

I'm not sure if it's feasible. I can try to look into it, but in any case, I believe that even if it is possible today, it should probably be handled in a separate PR, right?

> If so, is it worth complicating the nmstatectl flow just to avoid creating redundant minimal ISOs?
No strong opinion, but something to consider:)

I think the updated flow isn’t just for s390x, it also introduces caching for nmstatectl, which I find pretty nice. In fact, it feels like an improvement over the previous approach. It might involve a bit more code, but I think it's a more logical and correct solution overall. That said, it's probably best to wait and hear @carbonin ’s opinion on it too.

carbonin (commented):
Approach looks good, just some things that I think would be good to tidy up.

pkg/imagestore/imagestore.go
```diff
@@ -314,6 +320,65 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
+
+	// Extract ISO to temp directory
+	if err = isoeditor.Extract(fullPath, tempDir); err != nil {
+		return fmt.Errorf("failed to extract ISO: %v", err)
+	}
+
+	// Extract nmstatectl from rootfs and cache it
+	rootfsPath := filepath.Join(tempDir, "images/pxeboot/rootfs.img")
+
+	nmstateHandler := s.isoEditor.GetNmstateHandler()
```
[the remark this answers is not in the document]
linoyaslan: `	is, err := imagestore.NewImageStore(
		isoeditor.NewEditor(Options.DataTempDir, isoeditor.NewNmstateHandler(Options.DataTempDir, &isoeditor.CommonExecuter{})),
		Options.DataDir,
		Options.ImageServiceBaseURL,
		Options.InsecureSkipVerify,
		versions,
		Options.OSImageDownloadTrustedCAFile,
		osImageDownloadHeadersMap,
		osImageDownloadQueryParamsMap)`

`type rhcosEditor struct {
	workDir        string
	nmstateHandler NmstateHandler
}`


The reason is that nmstateHandler is already included in rhcosEditor. If you think it would be better to pass it through imageStore instead I can do that, but could you please explain the reasoning behind the suggestion?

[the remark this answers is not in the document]
carbonin: The reasoning is just keeping isoEditor public API focused on editing ISOs.

If the image store has a dependency on the nmstate handler then it should take in that handler as a dependency, just like anything else.

pkg/isoeditor/nmstate_handler.go
```diff
@@ -42,13 +64,30 @@ func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatect
 		return err
 	}
 
-	// write for caching
+	return nil
+}
+
+func (n *nmstateHandler) ExtractAndCacheNmstatectl(rootfsPath, nmstatectlPath string) error {
```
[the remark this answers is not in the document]
carbonin: Why can't `CreateNmstateRamDisk` just be this? They do exactly the same thing.

```go
func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatectlPath string) error {

    if err := ExtractAndCacheNmstatectl(rootfsPath, nmstatectlPath); err != nil {
        return err
    }
    // copy nmstatectlPath to ramDiskPath
}
```

pkg/imagestore/imagestore.go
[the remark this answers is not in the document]
linoyaslan: I see. So I pushed new changes, which include:

1. Decoupled nmstateHandler from isoEditor.
2. Direct dependency injection into ImageStore.

I'd appreciate your re-review. Thank you!

pkg/isoeditor/rhcos_test.go
```diff
@@ -143,3 +143,23 @@ var _ = Context("with test files", func() {
 		})
 	})
 })
+
+var _ = Describe("Editor", func() {
```
carbonin, line 147: This doesn't seem to be doing anything. Why was it added?

main.go
```diff
@@ -112,14 +112,15 @@ func main() {
 	}
 
 	is, err := imagestore.NewImageStore(
-		isoeditor.NewEditor(Options.DataTempDir, isoeditor.NewNmstateHandler(Options.DataTempDir, &isoeditor.CommonExecuter{})),
+		isoeditor.NewEditor(Options.DataTempDir),
```
carbonin, line 115: Why change this?

pkg/imagestore/imagestore.go
```diff
@@ -314,6 +322,63 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
...
+	// Get the path to the full ISO
+	fullPath := s.PathForParams(ImageTypeFull, openshiftVersion, arch)
+
+	// Extract ISO to temp directory
+	if err = isoeditor.Extract(fullPath, tempDir); err != nil {
+		return fmt.Errorf("failed to extract ISO: %v", err)
+	}
+
+	// Extract nmstatectl from rootfs and cache it
+	rootfsPath := filepath.Join(tempDir, "images/pxeboot/rootfs.img")
```
carbonin, line 371: Should this be a constant somewhere? Maybe in isoeditor?

pkg/imagestore/imagestore.go
carbonin: Yeah I think creating minimal ISOs for Z is out of scope for this PR.
Additionally I never particularly liked the nmstate image creation being a side-effect of creating the minimal ISO. I prefer this approach as it splits those unrelated concepts

pkg/isoeditor/nmstate_handler_test.go
```diff
@@ -71,12 +72,188 @@ var _ = Context("with test files", func() {
...
+
+			err = nmstateHandler.CreateNmstateRamDisk("", ramDiskPath, nmstatectlPathForCaching)
 			Expect(err).ToNot(HaveOccurred())
 
 			exists, err := fileExists(ramDiskPath)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(exists).To(BeTrue())
 		})
+
+		It("uses cached nmstatectl when available", func() {
```
carbonin, line 88: This tests the caching so the previous test doesn't need the cached content added before the call.

pkg/isoeditor/nmstate_handler_test.go
```diff
@@ -71,12 +72,188 @@ var _ = Context("with test files", func() {
...
+			// This tests the error path in ExtractAndCacheNmstatectl line 56
+			_, err = nmstateHandler.ExtractAndCacheNmstatectl(rootfsPath, nmstatectlCachePath)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("failed to read cached nmstatectl"))
+		})
 	})
+
+	// Note: buildNmstateCpioArchive and extractNmstatectl are private functions,
+	// so we test them indirectly through ExtractAndCacheNmstatectl which provides
+	// comprehensive coverage of the extraction logic through various scenarios above.
```
carbonin, line 258: We probably don't need this comment

pkg/isoeditor/nmstate_handler.go
```diff
@@ -31,24 +32,62 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
 }
 
 func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatectlPath string) error {
```
carbonin, line 34: Don't think we need `rootfsPath` anymore, right?

pkg/isoeditor/nmstate_handler.go
```diff
@@ -31,24 +32,62 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
...
+	//	cachedContent, readErr := os.ReadFile(nmstatectlPath)
+	//	if readErr != nil {
+	//		return nil, fmt.Errorf("failed to read cached nmstatectl: %v", readErr)
+	//	}
+	//	return cachedContent, nil
+	//}
+
+	log.Debugf("Extracting and caching nmstatectl from %s to %s", rootfsPath, nmstatectlPath)
+
+	compressedCpio, err := n.buildNmstateCpioArchive(rootfsPath)
```
carbonin, line 79: How about we make this the public method here and move the write into the caller?

pkg/isoeditor/nmstate_handler.go
```diff
@@ -30,28 +31,66 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
...
+//	}
+//
+//	// Write to cache
+//	err = os.WriteFile(nmstatectlPath, compressedCpio, 0755) //nolint:gosec
+//	if err != nil {
+//		return err
+//	}
+//
+//	return nil
+//}
```
carbonin, line 91: This can all be removed

pkg/isoeditor/nmstate_handler.go
```diff
@@ -30,28 +31,66 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
...
 
-func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatectlPath string) error {
-	compressedCpio, err := n.buildNmstateCpioArchive(rootfsPath)
-	if err != nil {
-		return err
+func (n *nmstateHandler) CreateNmstateRamDisk(ramDiskPath, nmstatectlPath string) error {
+	//compressedCpio, err := n.ExtractAndCacheNmstatectl(rootfsPath, nmstatectlPath)
+	//if err != nil {
+	//	return err
+	//}
```
carbonin, line 38: Let's remove these commented lines if we're keeping this function

pkg/isoeditor/nmstate_handler_test.go
```diff
@@ -34,49 +35,220 @@ var _ = Context("with test files", func() {
...
+		//	Expect(err).NotTo(HaveOccurred())
+		//
+		//	newNmstateHandler = NewNmstateHandler(workDir, mockExecuter)
+		//
+		//	// The nmstatectl "file" now exists (as a directory) but reading it will fail
+		//	// This tests the error path in ExtractAndCacheNmstatectl line 56
+		//	err = newNmstateHandler.ExtractAndCacheNmstatectl(rootfsPath, nmstatectlCachePath)
+		//	Expect(err).To(HaveOccurred())
+		//	Expect(err.Error()).To(ContainSubstring("failed to read cached nmstatectl"))
+		//})
```
carbonin, line 252: Remove commented code

pkg/isoeditor/nmstate_handler.go
```diff
@@ -30,28 +31,66 @@ func NewNmstateHandler(workDir string, executer Executer) NmstateHandler {
 	}
 }
 
-func (n *nmstateHandler) CreateNmstateRamDisk(rootfsPath, ramDiskPath, nmstatectlPath string) error {
-	compressedCpio, err := n.buildNmstateCpioArchive(rootfsPath)
-	if err != nil {
-		return err
+func (n *nmstateHandler) CreateNmstateRamDisk(ramDiskPath, nmstatectlPath string) error {
```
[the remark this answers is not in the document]
linoyaslan: but we do:

```
	// write for RAM disk
	err := os.WriteFile(ramDiskPath, compressedCpio, 0755) //nolint:gosec
	if err != nil {
		return err
	}
```

If we remove this function, it would add more complexity to CreateMinimalISOTemplate. Personally, I prefer keeping this separation  unless you feel otherwise :)

As for the name. do you have a better suggestion that would better reflect its current behavior?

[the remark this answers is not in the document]
carbonin: Sure, but functionally this is just `cp -n nmstatectlPath ramDiskPath` right?
There's nothing specific about nmstate or ramdisks at all, this just copies files now.

So if we want to keep this as a separate function, then I'd just make a `copy(src, dst string)` helper function in `rhcos.go`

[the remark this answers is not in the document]
linoyaslan: Sorry for the long discussion on this tiny thing, I just want to make sure I understand. Do you mean simply renaming it to copy, or **implementing** a copy method as part of some interface?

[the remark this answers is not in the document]
carbonin: >but we are copying the nmstatectl binary to the correct path /images/nmstate.img

But this path is input to the function, there's nothing nmstate-specific about the implementation.

>Regarding your suggestion, do you mean simply renaming it to copy, or implementing a copy method as part of some interface?

I mean make this a non-public function called `copy` wherever it's called from. We don't need a public function called `CreateNmstateRamDisk` that we just use to copy a file.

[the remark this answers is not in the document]
linoyaslan: Done. I'd appreciate your re-review :)

pkg/imagestore/imagestore_test.go
```diff
@@ -518,6 +523,145 @@ var _ = Context("with a data directory configured", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+					Expect(os.WriteFile(nmstatectlPath, []byte("cached-nmstatectl"), 0755)).To(Succeed()) //nolint:gosec
+
+					// No CreateMinimalISOTemplate should be called for s390x
+					mockEditor.EXPECT().CreateMinimalISOTemplate(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
+
+					Expect(is.Populate(ctx)).To(Succeed())
+
+					// Verify nmstatectl cache exists
+					_, err = os.Stat(nmstatectlPath)
```
carbonin, line 560: Isn't this the same path that this test wrote the file to earlier? Shouldn't we be checking some other path here to ensure the file was copied? Otherwise we're only testing the case where the file is already present.

I'd expect a mock for getting the file from the ISO and for getting the cpio form the rootfs.

pkg/imagestore/imagestore_test.go
```diff
@@ -518,6 +523,145 @@ var _ = Context("with a data directory configured", func() {
...
+						// Mock minimal ISO creation
+						expectedRootfs := fmt.Sprintf("http://images.example.com/boot-artifacts/rootfs?arch=%s&version=%s", arch, archVersion["openshift_version"])
+						mockEditor.EXPECT().CreateMinimalISOTemplate(gomock.Any(), expectedRootfs, arch, gomock.Any(), archVersion["openshift_version"], nmstatectlPath).Return(nil)
+
+						Expect(is.Populate(ctx)).To(Succeed())
+
+						// Verify nmstatectl cache exists
+						content, err := os.ReadFile(nmstatectlPath)
+						Expect(err).NotTo(HaveOccurred())
+						Expect(content).To(Equal([]byte(cacheContent)))
```
carbonin, line 602: Same here, we're expecting the file we wrote to be present. The only thing these are testing is that we didn't delete the file by calling `Populate`.

pkg/imagestore/imagestore_test.go
linoyaslan: Yeah, so the test currently only check the cache check:

```
// Check if nmstatectl is already cached
	if _, err = os.Stat(nmstatectlPath); err == nil {
		log.Debugf("nmstatectl already cached for %s-%s", openshiftVersion, arch)
		return nil
	}
```

in extractAndCacheNmstatectl and doesn’t cover GetFileFromISO or BuildNmstateCpioArchive. That’s the same limitation we discussed when I used the Extract method. Since the GetFileFromISO  not mocked, I’m unsure how best to handle it. We could wrap it behind an interface or something and generate a mock, what do you think?

carbonin: Ah right, forgot about that.

How about we add something to integration tests to verify? At least we should have some versions in there that are supported for nmstate so we pull it out on startup.

linoyaslan:
@carbonin I'd appreciate your re-review :) Thanks!

integration_test/images_test.go
```diff
@@ -168,26 +182,52 @@ var _ = Describe("Image integration tests", func() {
...
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/files", imageID), "file_name=discovery.ign"),
+						ghttp.RespondWith(http.StatusOK, tc.expectedIgnition),
+					),
+				)
+				// 5) pxe-initrd path: minimal-initrd always requested; 200 when provided, else 204
+				if len(tc.expectedRamdisk) > 0 {
+					assistedServer.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("GET", fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/minimal-initrd", imageID)),
```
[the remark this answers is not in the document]
linoyaslan: mmm iiuc,  when you call assistedServer.AppendHandlers(), you're adding handlers to a queue. Each incoming HTTP request consumes the next handler in the queue. If the actual request sequence doesn't match the queued handler sequence, it will fail with "Unhandled request" (this happened to me already) .so we must call them twice, unless there is another way to do that ?

integration_test/images_test.go
```diff
@@ -235,20 +275,38 @@ var _ = Describe("Image integration tests", func() {
...
+					defer rc.Close()
+
+					got, err := readIgnitionContentFromGzCpio(rc)
 					Expect(err).NotTo(HaveOccurred())
-					Expect(hdr.Name).To(Equal("config.ign"))
-					Expect(hdr.Size).To(Equal(int64(len(tc.expectedIgnition))))
-					content, err := io.ReadAll(cpioReader)
+					Expect(got).To(Equal(tc.expectedIgnition))
+
+					By("verifying pxe-initrd includes nmstate archive for >= 4.18 when ramdisk is present")
```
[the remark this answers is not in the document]
linoyaslan: following your other comment above, if I split the nmstate test out, each It() gets its own BeforeEach setup and a fresh handler queue, right? The nmstate-specific test only calls the /pxe-initrd endpoint, but our shared BeforeEach enqueues the other handlers first, which cause issues to my nmstate test. to solve that, I can move it into a separate Context with its own BeforeEach if you think it's worth it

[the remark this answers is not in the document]
carbonin: Correct, each `It` runs `BeforeEach`.

I would say that these test cases are ISO and ISO-type specific which we actually don't care about for what you changed.
Maybe this should be totally separate?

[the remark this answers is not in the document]
carbonin: Yeah, I don't think looping over the same list is too big a deal

PR #472, opened 2025-08-16
OCPBUGS-37943: FIPS based ISO Boot support for s390x arch

phani2898:
### Summary
This PR introduces enhancements to the `kargs.go` file, particularly focusing on appending kargs into the boot image itself where ignition is already embedded using the staging ISO boot image files. This is necessary for architecture like s390x where ignition and kargs lies in the same boot image (`cdboot.img`) unlike x86 where we have different boot files for booth. 

It includes a newly added function for embedding kernel arguments into boot images `EmbedKargsIntoBootImage`, plus unit tests for the same to ensure reliability and correctness.

#### EmbedKargsIntoBootImage Function

**Purpose:** 

- Allows embedding additional kernel arguments (kargs) into boot images where ignition is also present in the same at a different offset.
- Fetches the kargs details such as default kargs, kargs size and offset from `coreos/kargs.jso` instead of calculating it based on the marker `COREOS_KARGS_EMBED_AREA` which is not present in s390x ISO boot files

**Benefit:** Enhances flexibility and control over boot-time configurations—crucial for architectures needing to embed the kernel parameters into ignition boot images (e.g., s390x).

**Rationale:** This is created as a public function to make sure it can be accessed from the installer repo to have decision making in the `appendKargs` function to call this function instead NewKargsReader function, which relies on several factors which are not relevant for s390x and also archs having to embed kargs into same boot image as ignition content.

### Unit Tests for EmbedKargsIntoBootImage

The tests simulate real ISO staging scenarios by creating temporary base and staging directories, then exercising both failure cases and success cases.

**Failure Scenarios Covered:**

- Missing config file: Fails when /coreos/kargs.json cannot be read from the base ISO path.
- Malformed JSON: Fails when kargs.json contains invalid JSON.
- Empty entries: Fails when no files entries are present in kargs.json.
- Missing staging files: Fails if a listed target file does not exist in the staging directory.
- Field size exceeded (explicit size): Errors if combined default + custom kernel args exceed the configured size field.
- Field size exceeded (inferred size): Errors if size=0 but the available file space (based on length/offset) is insufficient.
- Invalid target file: Errors if the target path exists but is not a writable file (e.g., a directory).

**Success Scenario:**

- Multiple files with offsets: Verifies that custom kernel arguments are correctly embedded into two different files at different offsets, appended immediately after the default arguments.

### Functional tests

- Successfully generated Agent ISO image with `fips: true` having s390x arch
- Successfully created an SNO cluster with the same ISO image where FIPS is enabled after boot
- Performed the same for `fips: false` as well

## Assignees

/cc @zaneb 
/cc @andfasano 

## Links 

- This PR works in favour of s390x arch which gets consumed by `appendKargs` function in a decision making in the `installer` codebase. 
- Check out this PR: https://github.com/openshift/installer/pull/9892
- Corresponding slack discussion : https://redhat-internal.slack.com/archives/C06A57N4WEM/p1745332383117689

## Checklist

### DIFF
--- a/pkg/isoeditor/kargs.go
+++ b/pkg/isoeditor/kargs.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"regexp"
 
 	"github.com/openshift/assisted-image-service/pkg/overlay"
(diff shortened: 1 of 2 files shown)


zaneb:
Based on previous discussions, the issues preventing the existing `NewKargsReader()` API from working with s390 were:
* The [`kargsEmbedAreaBoundariesFinder()`](https://github.com/openshift/assisted-image-service/blob/1b29aba9d5bee83a536e40233a8170dbb2c72520/pkg/isoeditor/kargs.go#L104-L121) needs to use the `Offset`/`Size` fields in `coreos/kargs.json` to locate the embed area, and only fall back to the regexp search when `kargs.json` is not present.
* `NewKargsReader()` itself [assumes that the args end with a newline](https://github.com/openshift/assisted-image-service/blob/1b29aba9d5bee83a536e40233a8170dbb2c72520/pkg/isoeditor/kargs.go#L79-L82) instead of using the `End` field from `kargs.json` 
* Both `NewKargsReader()` and `NewRHCOSStreamReader()` stream the original files from the ISO with their changes overlaid, but when both modify the same file (as in the case of s390) this means that only one set of changes get applied.

I was under the impression that we had agreed that the solution to this was to have an API that combined the functionality of both `NewKargsReader()` and `NewRHCOSStreamReader()` so that both lots of changes could be overlaid on a single file. In my opinion this would be superior to creating an entirely new API that doesn't work like any of the existing ones so that consumers have to choose which to call based on platform.

phani2898:
> Agree with @zaneb. Let's make the existing implementation work for changing s390x kargs using the json file as input.
> 
> Also this approach wouldn't work for SaaS and I think we want this to work the same for ABI and for clusters created in console.redhat.com.

okay sure, will change accordingly. I introduced this new API because we may have to change few input parameters in the existing NewKargsReader to persist both kargs and ignition together in the same cdboot.img file

phani2898:
> Based on previous discussions, the issues preventing the existing `NewKargsReader()` API from working with s390 were:
> 
> * The [`kargsEmbedAreaBoundariesFinder()`](https://github.com/openshift/assisted-image-service/blob/1b29aba9d5bee83a536e40233a8170dbb2c72520/pkg/isoeditor/kargs.go#L104-L121) needs to use the `Offset`/`Size` fields in `coreos/kargs.json` to locate the embed area, and only fall back to the regexp search when `kargs.json` is not present.
> * `NewKargsReader()` itself [assumes that the args end with a newline](https://github.com/openshift/assisted-image-service/blob/1b29aba9d5bee83a536e40233a8170dbb2c72520/pkg/isoeditor/kargs.go#L79-L82) instead of using the `End` field from `kargs.json`
> * Both `NewKargsReader()` and `NewRHCOSStreamReader()` stream the original files from the ISO with their changes overlaid, but when both modify the same file (as in the case of s390) this means that only one set of changes get applied.
> 
> I was under the impression that we had agreed that the solution to this was to have an API that combined the functionality of both `NewKargsReader()` and `NewRHCOSStreamReader()` so that both lots of changes could be overlaid on a single file. In my opinion this would be superior to creating an entirely new API that doesn't work like any of the existing ones so that consumers have to choose which to call based on platform.

zaneb:
> even if this logic fetching kargs from kargs.json is fixed still this will not work because both ignition content and kargs are present in the same image cdboot.img for s390x.

Right, this was captured by the 3rd point:

> > * Both NewKargsReader() and NewRHCOSStreamReader() stream the original files from the ISO with their changes overlaid, but when both modify the same file (as in the case of s390) this means that only one set of changes get applied.

which was why I suggested:

> > to have an API that combined the functionality of both NewKargsReader() and NewRHCOSStreamReader() so that both lots of changes could be overlaid on a single file.

PR #474, opened 2025-08-17
Allow empty osimage list when qcow2 image is used

jianzzha:
## Description

The OpenShift assisted CAPI provider uses qocw2 image to bootstrap baremetal servers. In this provision path, there is no need to download iso images, though assisted image service is still started.

The related PR in the assisted service repo is:  https://github.com/openshift/assisted-service/pull/7940

## Assignees

/cc @
/cc @

## Checklist


jianzzha:
assisted image service is not needed to run with this approach: https://github.com/openshift/assisted-service/compare/master...rccrdpccl:assisted-service:allow-disabling-image-service

PR #477, opened 2025-08-19
MGMT-21574: Allow `rhel-coreos-` as a valid volume ID prefix

jlebon:
## Description

RHCOS 9.6/10.1 ISOs will soon have that volume ID there. The change was done thinking it was low risk since it's mostly cosmetic but it turns out we check for this here.

We could revert it, but... hopefully this patch is enough to fix things.

## How was this code tested?

`make`

## Assignees

/cc @carbonin 

## Links

- https://github.com/osbuild/osbuild/pull/2148
- https://redhat-internal.slack.com/archives/C999USB0D/p1755549666133839?thread_ts=1755284973.982289&cid=C999USB0D

## Checklist

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -231,11 +231,21 @@ func validateISOID(path string) error {
 		return err
 	}
 
-	if !strings.HasPrefix(volumeID, "rhcos-") && !strings.HasPrefix(volumeID, "fedora-coreos-") && !strings.HasPrefix(volumeID, "scos-") {
+	if !isoIDHasValidPrefix(volumeID) {
 		return fmt.Errorf("ISO volume identifier (%s) is invalid", volumeID)
 	}
 
 	return nil
+}
+
+func isoIDHasValidPrefix(volumeID string) bool {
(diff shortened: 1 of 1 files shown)


carbonin:
If we have a published image with this prefix available can you add it to the list in integration_test/images_test.go?

carbonin:
Also if this is something we specifically shouldn't be testing for as a way to identify these ISOs, maybe we should just remove it or switch to checking something more reliable?

jlebon:
> > But it should get there in the next bootimage bump
> 
> What does this mean? Is this a "real" release or some kind of prerelease? I'm just trying to sort out how quickly we need this out there for existing users.

It will eventually land in both 4.20 and 4.19 as we bump their respective bootimage references in `rhcos.json` in the installer.

carbonin (approved):

carbonin:
>Should it be merged anyway?

I think I'd rather look into a more stable way of ensuring the ISO is something we can work with.
Maybe we check for what we actually use (ignition info file or ignition file where we expect it to be)?
Or maybe there's some other detail we can look for?

PR #827, opened 2026-04-19
AGENT-1497: Rename OVE ISO download filename to agent-ove.<arch>.iso

omer-vishlitzky:
## Summary
- OVE/disconnected ISOs were downloaded as `<cluster_id>-discovery.iso`, which is confusing for users
- Renamed to `agent-ove.<arch>.iso` (e.g., `agent-ove.x86_64.iso`) for the disconnected-iso type only
- Regular full/minimal ISOs retain existing `<id>-discovery.iso` naming

## Test plan
- [x] Unit test added verifying `Content-Disposition` header contains `agent-ove.x86_64.iso` for disconnected-iso type
- [x] All existing handler tests pass (91/91)
- [ ] Verify OVE ISO download in SaaS UI names file correctly

Jira: https://issues.redhat.com/browse/AGENT-1497

### DIFF
--- a/internal/handlers/iso.go
+++ b/internal/handlers/iso.go
@@ -84,6 +84,9 @@ func (h *isoHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer isoReader.Close()
 
 	fileName := fmt.Sprintf("%s-discovery.iso", params.imageID)
+	if params.imageType == imagestore.ImageTypeDisconnectedIso {
+		fileName = fmt.Sprintf("agent-ove.%s.iso", params.arch)
+	}
 	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s", fileName))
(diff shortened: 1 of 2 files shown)


internal/handlers/iso.go
```diff
@@ -83,7 +83,12 @@ func (h *isoHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	defer isoReader.Close()
 
-	fileName := fmt.Sprintf("%s-discovery.iso", params.imageID)
+	var fileName string
+	if params.imageType == imagestore.ImageTypeDisconnectedIso {
+		fileName = fmt.Sprintf("agent-ove.%s.iso", params.arch)
+	} else {
+		fileName = fmt.Sprintf("%s-discovery.iso", params.imageID)
+	}
```
rccrdpccl: Can we remove the else condition? This coudl be the default value, and override it if OVE

PR #831, opened 2026-04-21
ACM-33009: Fix crash during erofs nmstate extraction

carbonin:
## Description

Current (4.22) ISOs with erofs type root filesystems cause a crash while trying to find the nmstatectl binary. This is due to a bug in the go-erofs library that was fixed
(https://github.com/erofs/go-erofs/pull/23), but even after the fix the library doesn't support the compression strategy this filesystem uses leading to errors like "unsupported incompatible feature 0x2: not implemented".

Using dump.erofs directly for searching for the file as well as extracting it is the strategy that is least likely to break as the dump.erofs utility is tied directly to the reference implementation of the filesystem spec.

## How was this code tested?

I reproduced the issue using the latest image service container image in an ACM context with the RHCOS ISO listed in the bug report. I built the fixed image and deployed it in the same environment and was able to register an agent with the problematic minimal ISO with static networking configured. Also the crashes stopped.

## Assignees

/cc @linoyaslan 
/cc @giladravid16 

## Links

Resolves https://redhat.atlassian.net/browse/ACM-33009

## Checklist


linoyaslan:
[@coderabbitai](https://github.com/coderabbitai) review

pkg/isoeditor/nmstate_handler.go
```diff
@@ -252,3 +227,54 @@ func (e *erofsExtractor) ExtractNmstatectl(nmstateDir string) (string, error) {
 
 	return "nmstatectl", nil
 }
+
+// findFileInErofs recursively searches for a file in an EROFS image using dump.erofs
+func (e *erofsExtractor) findFileInErofs(nmstateDir, dirPath, filename string) (string, error) {
+	output, err := e.executer.Execute(fmt.Sprintf("dump.erofs --path=%s --ls root.erofs", dirPath), nmstateDir)
```
linoyaslan, line 233: I know `Execute` is a function wrote a while ago, but I'm wondering about the general security risks here using `exec.Command("bash", "-c", ...)` which can be dangerous, no? I mean it allows shell features like pipes and other special characters?

linoyaslan:
Looks good to me, I just put a small comment about some concern that isn't directly related to this PR, but I'm wondering if it could be addressed here as well, if you agree :)

pkg/isoeditor/nmstate_handler.go
carbonin: It's something that we should pay attention to, but it should be fine since none of this data is ever coming from a user as far as I can tell. It is definitely something we should check for injection risks if it were running as a part of a user request for example.

Regarding the shell features .. Yeah I think it was done that way because existing callers are effectively exploiting this with calls like these:

```
pkg/isoeditor/nmstate_handler.go:186:   list, err := e.executer.Execute(fmt.Sprintf("%s ; unsquashfs -dest '' -lc %s", ulimit, "root.squashfs"), nmstateDir)
pkg/isoeditor/nmstate_handler.go:199:   _, err = e.executer.Execute(fmt.Sprintf("%s ; unsquashfs -no-xattrs %s -extract-file %s", ulimit, "root.squashfs", binaryPath), nmstateDir)
```

I imagine that's why it was written this way

linoyaslan: Thanks for the explanation Nick!

PR #854, opened 2026-05-19
MGMT-24427: Validate boot artifacts in ISO after download

asadawar:
## Summary

https://redhat.atlassian.net/browse/MGMT-24427

After downloading an ISO during `Populate()`, validate that expected boot artifacts (`rootfs.img`, `vmlinuz`) exist inside the ISO and are above minimum size thresholds. This extends the existing post-download validation (`validateISOID`) with boot artifact integrity checks.

### What it does
- Adds `validateBootArtifacts(isoPath, arch)` -- checks rootfs (>100 MB) and kernel (>5 MB) exist in the ISO using `GetISOFileInfo` (metadata-only, no file content read)
- Integrates into `Populate()` right after `validateISOID()` -- same pattern for size validation errors (remove ISO, return error for re-download on next run)
- Handles s390x `kernel.img` fallback (matching existing handler logic in `boot_artifacts.go`)
- Differentiates error types: validation failures (missing/undersized artifacts) trigger file removal for re-download; generic errors are logged without deleting potentially valid ISOs

### What it catches
- Truncated ISO downloads where boot artifacts are undersized (e.g. CDN/proxy buffering failures)
- Corrupt ISOs where expected boot artifacts are missing entirely
- ISOs with different internal layouts (e.g. new OCP versions)

### Design decisions
- **Typed validation errors** (`bootArtifactValidationError`) -- missing or undersized boot artifacts are treated as definitive corruption, triggering file deletion and re-download. Generic read errors are logged but don't delete the file.
- **Conservative thresholds** -- 100 MB rootfs / 5 MB kernel, ~10x below real artifact sizes (rootfs ~1.1 GB, kernel ~13 MB on mirror.openshift.com), to avoid false positives across releases
- **Real ISOs in tests** -- Populate tests use `genisoimage` to create ISOs with valid boot artifacts, so `validateBootArtifacts` runs against actual ISO content with no test stubs

## Test plan
- [x] 7 unit tests for `validateBootArtifacts` using real ISOs (via `genisoimage`)
  - Valid ISO succeeds
  - Missing rootfs fails
  - Undersized rootfs fails
  - Missing kernel fails
  - Undersized kernel fails
  - s390x `kernel.img` fallback succeeds
  - s390x with both kernel types missing fails
- [x] Populate test for boot artifact failure (undersized rootfs ISO, verifies file deletion)
- [x] All 40 tests pass
- [x] `go vet` clean, `gofmt` clean

## Summary by CodeRabbit

* **New Features**
  * ISO validation now ensures required boot artifacts (rootfs and kernel) meet minimum size constraints; includes s390x-specific kernel fallback logic.

* **Bug Fixes**
  * Partially downloaded or invalid ISO files are removed when boot-artifact validation fails.
  * Non-size validation issues are logged as warnings and no longer always fail the download.

* **Tests**
  * Tests expanded to generate and verify ISOs with presence/size and platform-specific kernel scenarios.

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -90,6 +90,13 @@ const (
 	ImageTypeFull            = "full-iso"
 	ImageTypeMinimal         = "minimal-iso"
 	ImageTypeDisconnectedIso = "disconnected-iso"
+
+	// Minimum expected sizes for boot artifacts in RHCOS ISOs.
+	// Reference: RHCOS artifacts on mirror.openshift.com — rootfs is ~1.1 GB, kernel is ~13 MB
+	// for a typical x86_64 release. Thresholds are set ~10x below real sizes to catch
+	// truncated downloads without false-positiving on normal variance across releases.
+	minRootfsSize = 100 * 1024 * 1024
+	minKernelSize = 5 * 1024 * 1024
 )
 
 func NewImageStore(ed isoeditor.Editor, dataDir, imageServiceBaseURL string, insecureSkipVerify bool, versions []map[string]string,
@@ -233,6 +240,40 @@ func validateISOID(path string) error {
 	return nil
 }
 
+type bootArtifactValidationError struct {
+	message string
+}
+
+func (e *bootArtifactValidationError) Error() string { return e.message }
+
+// validateBootArtifacts checks that the downloaded ISO contains rootfs and kernel
+// above minimum expected sizes. This catches rare cases where a download is truncated
+// by a CDN or proxy, producing an ISO that passes volume ID validation but fails to boot.
+func validateBootArtifacts(isoPath, arch string) error {
+	_, rootfsSize, err := isoeditor.GetISOFileInfo(isoeditor.RootfsImagePath, isoPath)
+	if err != nil {
+		return &bootArtifactValidationError{fmt.Sprintf("rootfs not found in ISO: %v", err)}
+	}
+	if rootfsSize < minRootfsSize {
+		return &bootArtifactValidationError{fmt.Sprintf("rootfs size %d bytes is below minimum %d", rootfsSize, minRootfsSize)}
+	}
+
+	kernelPath := "/images/pxeboot/vmlinuz"
+	_, kernelSize, err := isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	if err != nil && arch == "s390x" {
+		kernelPath = "/images/pxeboot/kernel.img"
+		_, kernelSize, err = isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	}
+	if err != nil {
+		return &bootArtifactValidationError{fmt.Sprintf("kernel not found in ISO: %v", err)}
+	}
+	if kernelSize < minKernelSize {
+		return &bootArtifactValidationError{fmt.Sprintf("kernel size %d bytes is below minimum %d", kernelSize, minKernelSize)}
+	}
+
+	return nil
+}
+
 func (s *rhcosStore) Populate(ctx context.Context) error {
 	if err := s.cleanDataDir(); err != nil {
 		return err
@@ -269,6 +310,18 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 					}
 					log.Error(message)
 					return errors.New(message)
+				}
+				if err := validateBootArtifacts(fullPath, arch); err != nil {
+					var sizeErr *bootArtifactValidationError
+					if errors.As(err, &sizeErr) {
+						message := fmt.Sprintf("failed to validate boot artifacts in %s: %v", fullPath, err)
+						if removeErr := os.Remove(fullPath); removeErr != nil {
+							log.WithError(removeErr).Errorf("failed to remove corrupt ISO %s", fullPath)
+						}
+						log.Error(message)
+						return errors.New(message)
+					}
+					log.WithError(err).Warningf("boot artifact validation could not complete for %s", fullPath)
 				}
 			}
 
--- a/pkg/imagestore/imagestore_test.go
+++ b/pkg/imagestore/imagestore_test.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"testing"
@@ -136,6 +137,7 @@ var _ = Context("with a data directory configured", func() {
 				ctrl               *gomock.Controller
 				mockEditor         *isoeditor.MockEditor
 				mockNmstateHandler *isoeditor.MockNmstateHandler
+				isoDir             string
 				validVolumeID      = "rhcos-411.86.202210041459-0"
 				version            = map[string]string{
 					"openshift_version": "4.8",
@@ -145,24 +147,32 @@ var _ = Context("with a data directory configured", func() {
 			)
 
 			BeforeEach(func() {
+				requireGenisoimage()
 				ctrl = gomock.NewController(GinkgoT())
 				mockEditor = isoeditor.NewMockEditor(ctrl)
 				mockNmstateHandler = isoeditor.NewMockNmstateHandler(ctrl)
 				osImageDownloadHeadersMap = map[string]string{}
 				osImageDownloadQueryParamsMap = map[string]string{}
+				var err error
+				isoDir, err = os.MkdirTemp("", "populateTestISO")
+				Expect(err).NotTo(HaveOccurred())
 			})
 
-			isoInfo := func(id string) ([]byte, http.Header) {
-				content := make([]byte, 32840)
-				copy(content[32808:], id)
+			AfterEach(func() {
+				os.RemoveAll(isoDir)
+			})
+
+			readTestISO := func(volumeID string) ([]byte, http.Header) {
+				isoFile := buildTestISO(isoDir, volumeID, minRootfsSize+1024, minKernelSize+1024)
+				content, err := os.ReadFile(isoFile)
+				Expect(err).NotTo(HaveOccurred())
 				header := http.Header{}
 				header.Add("Content-Length", strconv.Itoa(len(content)))
-
 				return content, header
 			}
 
 			It("downloads an image correctly", func() {
-				isoContent, isoHeader := isoInfo(validVolumeID)
+				isoContent, isoHeader := readTestISO(validVolumeID)
 				ts.AppendHandlers(
 					ghttp.CombineHandlers(
 						ghttp.VerifyRequest("GET", "/some.iso"),
@@ -211,6 +221,7 @@ var _ = Context("with a data directory configured", func() {
 				ctrl               *gomock.Controller
 				mockEditor         *isoeditor.MockEditor
 				mockNmstateHandler *isoeditor.MockNmstateHandler
+				isoDir             string
 				validVolumeID      = "rhcos-411.86.202210041459-0"
 				version            = map[string]string{
 					"openshift_version": "4.8",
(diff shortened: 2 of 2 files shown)


pkg/imagestore/imagestore_test.go
```diff
@@ -757,6 +768,94 @@ var _ = Describe("NewImageStore", func() {
...
+		}
+		if includeKernelImg {
+			kernel := make([]byte, kernelSize)
+			Expect(os.WriteFile(filepath.Join(filesDir, "images/pxeboot/kernel.img"), kernel, 0600)).To(Succeed())
+		}
+
+		isoFile := filepath.Join(isoDir, "test.iso")
+		cmd := exec.Command("genisoimage", "-rational-rock", "-J", "-joliet-long",
+			"-V", "rhcos-test", "-o", isoFile, filesDir)
+		Expect(cmd.Run()).To(Succeed())
```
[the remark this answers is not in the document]
asadawar: Fixed — added LookPath check with Skip and CombinedOutput for diagnostics.

asadawar:
**Re: genisoimage guard** — Good catch, fixed. Added `exec.LookPath("genisoimage")` with `Skip()` in `BeforeEach`, and switched to `cmd.CombinedOutput()` for better failure diagnostics.

**Re: Truncate instead of byte slices** — Good catch, fixed. Using `file.Truncate()` to create sparse files instead of allocating 100+ MB byte slices in memory.

**Re: s390x fallback on undersized vmlinuz** — Intentionally not changed. The current logic mirrors the handler in `boot_artifacts.go` exactly — fallback to `kernel.img` only when `vmlinuz` is absent, not undersized. A too-small `vmlinuz` means the ISO is corrupt, not that we should try an alternative kernel path.

pkg/imagestore/imagestore.go
```diff
@@ -233,6 +236,33 @@ func validateISOID(path string) error {
...
+	_, kernelSize, err := isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	if err != nil && arch == "s390x" {
+		kernelPath = "/images/pxeboot/kernel.img"
+		_, kernelSize, err = isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	}
+	if err != nil {
+		return fmt.Errorf("kernel not found in ISO: %v", err)
+	}
+	if kernelSize < minKernelSize {
+		return fmt.Errorf("kernel size %d bytes is below minimum %d", kernelSize, minKernelSize)
```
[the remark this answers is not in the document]
asadawar: Intentionally not changed. The current logic mirrors the handler in `boot_artifacts.go` (line 80) exactly — fallback to `kernel.img` only when `vmlinuz` is absent, not undersized. A too-small `vmlinuz` indicates ISO corruption, not a naming difference.

asadawar:
**Codecov feedback addressed:** Added a test for the `Populate()` error path when boot artifact validation fails. This covers the 6 lines Codecov flagged — the error handling block that removes the ISO and returns an error when `validateBootArtifactsFn` reports a failure.

New test: `"fails and removes the file when boot artifact validation fails"` — sets `validateBootArtifactsFn` to return an error, verifies `Populate()` fails with the expected message and the ISO file is cleaned up.

pkg/imagestore/imagestore.go
```diff
@@ -90,6 +90,9 @@ const (
 	ImageTypeFull            = "full-iso"
 	ImageTypeMinimal         = "minimal-iso"
 	ImageTypeDisconnectedIso = "disconnected-iso"
+
+	minRootfsSize = 100 * 1024 * 1024 // 100 MB — real rootfs is ~500 MB+
```
danielerez: Is there any reference for these values that maybe can be noted here?

pkg/imagestore/imagestore.go
```diff
@@ -233,6 +236,33 @@ func validateISOID(path string) error {
 	return nil
 }
 
+var validateBootArtifactsFn = validateBootArtifacts
```
danielerez: No need to create a stub just for tests. Would be nicer to reuse the createISO helper from Describe("validateBootArtifacts") in Populate download tests so downloaded ISOs include valid rootfs + kernel sizes. Then Populate can call validateBootArtifacts with no stub.

I.e. something like:
```
	createISO := func(includeRootfs bool, rootfsSize int, includeVmlinuz bool, kernelSize int, includeKernelImg bool) string {
		return buildTestISOFile(isoDir, testISOOptions{
			volumeID:         "rhcos-test",
			includeRootfs:    includeRootfs,
			rootfsSize:       rootfsSize,
			includeVmlinuz:   includeVmlinuz,
			kernelSize:       kernelSize,
			includeKernelImg: includeKernelImg,
		})
	}

	It("succeeds with valid boot artifacts", func() {
		isoFile := createISO(true, minRootfsSize+1024, true, minKernelSize+1024, false)
		Expect(validateBootArtifacts(isoFile, "x86_64")).To(Succeed())
	})
```

pkg/imagestore/imagestore.go
```diff
@@ -270,6 +300,14 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 					log.Error(message)
 					return errors.New(message)
 				}
+				if err := validateBootArtifactsFn(fullPath, arch); err != nil {
+					message := fmt.Sprintf("failed to validate boot artifacts in %s: %v", fullPath, err)
```
danielerez: We probably want to differentiate between a generic failure in the validateBootArtifacts function, and simply a failed validation (I.e. size is lower than expected).
I mean, we can basically only handle a scenario in which the file is corrupted and successfully deleted. Then, we would retry to download on next populate.
So something like: If validation failed (size is lower) and file was removed, then return an error. Otherwise, just log an error message.

pkg/imagestore/imagestore.go
```diff
@@ -233,6 +236,33 @@ func validateISOID(path string) error {
 	return nil
 }
 
+var validateBootArtifactsFn = validateBootArtifacts
+
+func validateBootArtifacts(isoPath, arch string) error {
```
danielerez, line 252: Worth adding some details in a comment here about the scenario we're tackling. I.e. Since it should be a rather rare situation to have a corrupt download, etc.

pkg/imagestore/imagestore.go
asadawar: Added a comment above the constants. Real rootfs is typically 500-900 MB, kernel is 10-12 MB. The thresholds are intentionally conservative to catch truncated downloads without false-positiving on size variance between releases.

pkg/imagestore/imagestore.go
asadawar: Good call. Removed the validateBootArtifactsFn indirection entirely. Extracted a shared buildTestISO helper that creates real ISOs via genisoimage, and the Populate tests now use it through a readTestISO wrapper. The mock HTTP server serves these real ISOs, so validateBootArtifacts runs against actual content during Populate. Added a requireGenisoimage() skip guard for environments where genisoimage is not available.

pkg/imagestore/imagestore.go
asadawar: Makes sense. Added a bootArtifactSizeError type for the size validation failures. In Populate, errors.As checks the type: if it is a size error (truncated/corrupt download), the file is deleted and an error is returned so the next populate retries. For generic failures (e.g. transient read error), we just log a warning and keep the file -- no point destroying a potentially valid ISO over a read hiccup.

pkg/imagestore/imagestore.go
asadawar: Added a doc comment on the function explaining the scenario: catches rare truncated downloads where the ISO passes volume ID validation but contains undersized boot artifacts that would fail at boot time.

pkg/imagestore/imagestore.go
```diff
@@ -233,6 +239,40 @@ func validateISOID(path string) error {
...
+	}
+
+	kernelPath := "/images/pxeboot/vmlinuz"
+	_, kernelSize, err := isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	if err != nil && arch == "s390x" {
+		kernelPath = "/images/pxeboot/kernel.img"
+		_, kernelSize, err = isoeditor.GetISOFileInfo(kernelPath, isoPath)
+	}
+	if err != nil {
+		return fmt.Errorf("kernel not found in ISO: %v", err)
```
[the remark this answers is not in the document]
asadawar: Good catch. A missing rootfs or kernel inside an RHCOS ISO is just as broken as an undersized one -- the ISO will never boot. Renamed to bootArtifactValidationError and now all four error paths (missing rootfs, undersized rootfs, missing kernel, undersized kernel) return the typed error, so Populate deletes the file and retries on next run. The generic fallback path is now only a safety net for unexpected future error types.

pkg/imagestore/imagestore_test.go
```diff
@@ -145,24 +147,32 @@ var _ = Context("with a data directory configured", func() {
...
-				copy(content[32808:], id)
+			readTestISO := func(volumeID string) ([]byte, http.Header) {
+				isoFile := buildTestISO(isoDir, volumeID, minRootfsSize+1024, minKernelSize+1024)
+				content, err := os.ReadFile(isoFile)
+				Expect(err).NotTo(HaveOccurred())
 				header := http.Header{}
 				header.Add("Content-Length", strconv.Itoa(len(content)))
-
 				return content, header
 			}
```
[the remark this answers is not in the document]
asadawar: The ISOs are created with truncateFile (sparse files), so genisoimage only writes metadata -- the actual file on disk is a few KB, not 100 MB. The full test suite runs in ~5.5 seconds. If this becomes a bottleneck we can add caching, but right now it is not a concern.

pkg/imagestore/imagestore.go
```diff
@@ -270,6 +310,18 @@ func (s *rhcosStore) Populate(ctx context.Context) error {
 					log.Error(message)
 					return errors.New(message)
 				}
+				if err := validateBootArtifacts(fullPath, arch); err != nil {
+					var sizeErr *bootArtifactValidationError
+					if errors.As(err, &sizeErr) {
+						message := fmt.Sprintf("failed to validate boot artifacts in %s: %v", fullPath, err)
+						if removeErr := os.Remove(fullPath); removeErr != nil {
+							log.WithError(removeErr).Errorf("failed to remove corrupt ISO %s", fullPath)
```
danielerez, line 319: Hmm, for later, probably worth considering a better handling for failed removal.
I.e. as it's not an ideal solution to just keep a corrupted file in store when failing to remove.
For example, we can try to rename and cleanup, etc.
But any way, not in the scope here, as we have a similar issue also for validateISOID:)

pkg/imagestore/imagestore.go
danielerez: Great, thanks. Any link maybe that can be a reference? I mean official docs or something?

danielerez:
Looks good. Just another nit comment:)

pkg/imagestore/imagestore.go
asadawar: Agreed, noted for a follow-up. Rename-then-cleanup would be safer. Same pattern exists in the validateISOID path above, so it would make sense to address both together.

pkg/imagestore/imagestore.go
asadawar: No official docs page for artifact sizes. The authoritative reference is the RHCOS mirror itself:

```
$ curl -sI https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.18/latest/rhcos-live-rootfs.x86_64.img | grep content-length
content-length: 1209448448   # ~1.1 GB

$ curl -sI https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.18/latest/rhcos-live-kernel-x86_64 | grep content-length
content-length: 13625464     # ~13 MB
```

Consistent across 4.14 through 4.22. Updated the code comment to reference mirror.openshift.com as the source.

asadawar:
@danielerez Thanks for the review and /lgtm! All CI checks are passing. Could you also /approve when you get a chance?

PR #979, opened 2026-07-23
MGMT-24907: Use openshift version to check if version is high enough for nmstate

giladravid16:
## Description

When implementing MGMT-24783, a part of initrd’s logic was missed.
The version that initrd receives can be an RHCOS version, but we still compare it to the minimum version allowed for nmstate, as if it’s always an openshift version.

When there are multiple OS Images with the same RHCOS version, there's no way to know exactly which openshift version the user originally wanted.
However, going with the highest version is the safest, since it only decides if we include the nmsatectl binary.
So picking a higher version than needed will result in including a binary that's not needed, picking a lower version than needed will result in a missing binary.

## Assignees

/cc @
/cc @

## Links

Closes [MGMT-24907](https://redhat.atlassian.net/browse/MGMT-24907)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved version handling when generating initrd overlays, ensuring nmstate content is included for supported RHCOS and OpenShift version combinations.
  * Requests with unsupported or unresolvable version and architecture combinations now return a clear `400 Bad Request` response instead of proceeding incorrectly.
* **Tests**
  * Added coverage for supported version mappings and nmstatectl inclusion in generated initrd responses.

### DIFF
--- a/internal/handlers/initrd.go
+++ b/internal/handlers/initrd.go
@@ -92,7 +92,12 @@ func initrdOverlayReader(imageStore imagestore.ImageStore, client *AssistedServi
 			return nil, "", http.StatusInternalServerError, fmt.Errorf("failed to create append reader for initrd: %v", err)
 		}
 
-		versionOK, err := common.VersionGreaterOrEqual(version, isoeditor.MinimalVersionForNmstatectl)
+		openshiftVersion, err := imageStore.OpenshiftVersionForParams(version, arch)
+		if err != nil {
+			return nil, "", http.StatusBadRequest, err
+		}
+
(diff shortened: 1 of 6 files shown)


carbonin:
What was special about this version (4.18)? Is it something that is a function of the image and so we should also be able to map it to an rhcos version? Or is it something about OCP itself?

cc @linoyaslan do you remember?

linoyaslan:
@carbonin iirc, the reason we decided to use the new flow starting from 4.18 is that older versions include old nmstate and we wanted to avoid useing the new flow from these versions to minimize risk in prod. The actual version guard should probably be 4.14 or 4.13 (I don't remember exactly which one) but iirc `nmstate.service` has been available on RHCOS since one of those versions only.

PR #1016, opened 2026-08-27
MGMT-25140: Add a struct for OS Image and correctly check if initrd should include the nmstatectl

giladravid16:
## Description

We currently treat an OS Image as a map from string to string, but on assisted-service we're going to include a non-string field to the OS Image struct which would break the existing code.

Also because the version initrd receives can be the RHCOS version we can't directly compare it with the minimum openshift version where we include nmstatectl. Instead we'll rely on the fact that the imagestore will only create the nmstatectl for OS Images with openshift versions that are at least the minimum. So in initrd, if nmstatectl exists for the version we'll include it.

## How was this code tested?

Deployed with RHCOS_VERSIONS including the new fields and checking that versions that should include nmstatectl include it while versions that shouldn't don't. 

## Assignees

/cc @
/cc @

## Links

Closes [MGMT-25140](https://redhat.atlassian.net/browse/MGMT-25140)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nmstate image detection by checking whether the required cached artifact exists.
  * Preserved clear server errors when nmstate image lookup fails.

* **Improvements**
  * Improved handling of OS image metadata, including image types and stream information.
  * Enhanced image cache lookup and cleanup behavior.

* **Tests**
  * Expanded coverage for typed OS image data, cache existence, stream metadata, and nmstate scenarios.

### DIFF
--- a/pkg/imagestore/imagestore.go
+++ b/pkg/imagestore/imagestore.go
@@ -23,48 +23,61 @@ import (
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
 )
 
-var DefaultVersions = []map[string]string{
+// OSImage describes an OS image entry from OS_IMAGES / RHCOS_VERSIONS JSON.
+// Field names and types match the assisted-service os-image schema so the shared
+// catalog JSON (including boolean default_os_stream) unmarshals correctly.
+type OSImage struct {
+	OpenshiftVersion string `json:"openshift_version"`
+	CPUArchitecture  string `json:"cpu_architecture"`
+	URL              string `json:"url"`
+	Version          string `json:"version"`
+	Type             string `json:"type,omitempty"`
+	OsStream         string `json:"os_stream,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/imagestore/imagestore.go
```diff
@@ -23,48 +23,61 @@ import (
...
+	OpenshiftVersion string `json:"openshift_version"`
+	CPUArchitecture  string `json:"cpu_architecture"`
+	URL              string `json:"url"`
+	Version          string `json:"version"`
+	Type             string `json:"type,omitempty"`
+	OsStream         string `json:"os_stream,omitempty"`
+	DefaultOsStream  *bool  `json:"default_os_stream,omitempty"`
+}
+
+var DefaultVersions = []OSImage{
```
carbonin, line 39: Not for this PR, but these are so old we should probably think about removing these defaults entirely. Surely no one is relying on them 🙏

carbonin (approved):

